PR #8, opened 2017-07-16
Whitelist navigation events

boundsj:
Also don't allow post if the events queue is empty.


boundsj:
@ericrwolfe for now, this change will avoid the need for us to make changes like https://github.com/mapbox/mapbox-events-ios/pull/7 for new nav events while the spec is being developed. It also patches up the issue where posts would be made with an empty array of events.

PR #10, opened 2017-09-20
Check if location background mode is enabled before setting it

frederoni:
Setting `self.locationManager.allowsBackgroundLocationUpdates = YES;` crashes if `location` is missing from `UIBackgroundModes` in the Info.plist.

Needs testing

@boundsj 👀


boundsj:
Thanks @frederoni! I proposed https://github.com/mapbox/mapbox-events-ios/pull/11 as an alternative and a slightly larger refactor. Closing here.

PR #11, opened 2017-09-21
Move background modes guard to location manager wrapper

boundsj:
The `allowsBackgroundLocationUpdates` API is not safe to use unless the host application has set location as a background mode. This guard was in place but was done at the level above where it should have been.

This change moves the guard to the MMECLLocationManagerWrapper that actually holds the real CLLocationManager instance. As the source of truth for all location manager instance interactions, it is the safest place to perform this check.

This avoids a crasher that could occur if the host app received only `in use` permissions from the user and did not have the `location` background mode in the app's plist and the client of this library set the `isMetricsEnabledForInUsePermissions` to `true` (a `false` value did not trigger the crash since setting CLLocationManager's `allowsBackgroundLocationUpdates` property to `false` (the default value) is not a "programmer error".

Subsumes https://github.com/mapbox/mapbox-events-ios/pull/10

### DIFF
--- a/MapboxMobileEvents/MMECLLocationManagerWrapper.h
+++ b/MapboxMobileEvents/MMECLLocationManagerWrapper.h
@@ -11,6 +11,7 @@
 @property (nonatomic) CLLocationAccuracy desiredAccuracy;
 @property (nonatomic) CLLocationDistance distanceFilter;
 @property (nonatomic, copy, readonly) NSSet<__kindof CLRegion *> *monitoredRegions;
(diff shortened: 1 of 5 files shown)


frederoni (approved):
Much better than the proposed change in #10 👍

captainbarbosa (approved):

boundsj:
Sorry for the delay. Now that https://github.com/mapbox/mapbox-events-ios/pull/13 has landed, I will be able to properly unit test this PR.

boundsj:
Tests added to catch the issue that caused this bug.

PR #15, opened 2017-09-29
Move test dependency to Cartfile.private

frederoni:
Cedar is only needed for testing this project so it should reside in Cartfile.private to avoid building Cedar in all apps making use of this library through Carthage.

@boundsj 👀

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,1 +1,1 @@
-github "pivotal/Cedar" ~> 1.0
+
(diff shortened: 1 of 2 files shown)


boundsj:
🙇

boundsj:
Tagged in https://github.com/mapbox/mapbox-events-ios/releases/tag/v0.2.11

Issue #12, opened 2017-09-25
Rewrite tests in Cedar


boundsj:
This project has used protocols and and custom fakes classes that implement those protocols to facilitate testing that requires integration with other classes the library provides and/or framework functionality. Although this has the benefit of not introducing any 3rd party testing frameworks as a project dependency, the downside is that not using a mocking framework makes the implementation more complex than it needs to be.

Since this library is written in Objective-C and because introducing a development dependency should not create any new issues with integration of this library with other SDKs and apps, we should switch to use the [Cedar](https://github.com/pivotal/cedar) testing framework. Cedar includes a test double API and also allows for a cleaner, BDD style syntax that'll make the tests in this project easier to write, read, and maintain.

rclee:
Cedar porting complete.

PR #17, opened 2017-10-18
Port API client tests to Cedar

rclee:
Probably missing something simple here. Added APIClientFake, spied on it, and am attempting to receive the postEvent method. Here is the current failure message I'm getting...

`FAILURE MMEAPIClient - Posting events when posting a single event post event should be successful
/Users/randalllee/workspace/mapbox-events-ios/MapboxMobileEventsTests/MMEAPIClientTests_cedar.mm:164 Expected <<MMEAPIClientFake: 0x7fdb471d19c0>> to have received message <postEvent:completionHandler:>, with arguments: <<MMEEvent name=location, attributes={
    applicationState = Foreground;
    event = location;
    operatingSystem = 1;
    sessionId = "instance-id-1";
    source = mapbox;
}>, anything> but received messages:
  setAccessToken:<access-token>
  setUserAgentBase:<user-agent-base>
  setHostSDKVersion:<host-sdk-version>`


boundsj:
@rclee this PR is about porting API client tests to Cedar. I updated the title to reflect that.

MapboxMobileEventsTests/MMEAPIClientTests_cedar.mm
```diff
@@ -0,0 +1,327 @@
...
+    
+    __block MMEAPIClient *apiClient;
+    
+    beforeEach(^{
+        apiClient = [[MMEAPIClient alloc] initWithAccessToken:@"access-token"
+                                                userAgentBase:@"user-agent-base"
+                                               hostSDKVersion:@"host-sdk-1"];
+    });
+    
+    describe(@"- testing API setup", ^{
```
boundsj: Maybe a better description would be "creating an api client instance"

MapboxMobileEventsTests/MMEAPIClientTests_cedar.mm
```diff
@@ -0,0 +1,327 @@
...
+                [NSUserDefaults standardUserDefaults] stub_method(@selector(objectForKey:)).with(MMETelemetryTestServerURL).and_return(nil);
+                
+                apiClient = [[MMEAPIClient alloc] initWithAccessToken:@"access-token"
+                                                        userAgentBase:@"user-agent-base"
+                                                       hostSDKVersion:@"host-sdk-1"];
+                
+                apiClient.sessionWrapper should_not be_nil;
+            });
+            
+            it(@"should equal the BaseURL constant", ^{
```
boundsj: "should set its base URL equal to the the base URL constant value"

MapboxMobileEventsTests/MMEAPIClientTests_cedar.mm
```diff
@@ -0,0 +1,327 @@
...
+            
+            beforeEach(^{
+                spy_on([NSUserDefaults standardUserDefaults]);
+                [NSUserDefaults standardUserDefaults] stub_method(@selector(objectForKey:)).with(MMETelemetryTestServerURL).and_return(nil);
+                
+                apiClient = [[MMEAPIClient alloc] initWithAccessToken:@"access-token"
+                                                        userAgentBase:@"user-agent-base"
+                                                       hostSDKVersion:@"host-sdk-1"];
+                
+                apiClient.sessionWrapper should_not be_nil;
```
boundsj: Consider moving this into an it block and maybe testing that it is the correct instance type

MapboxMobileEventsTests/MMEAPIClientTests_cedar.mm
```diff
@@ -0,0 +1,327 @@
...
+            });
+            
+            it(@"should equal the BaseURL constant", ^{
+                apiClient.baseURL should equal([NSURL URLWithString:MMEAPIClientBaseURL]);
+            });
+            
+            afterEach(^{
```
boundsj: nit: I think I prefer to have the afterEach block just below the beforeEach block at the top of the context so that you can more easily read about setup and teardown in one place. Not a big deal though.

MapboxMobileEventsTests/MMEAPIClientTests_cedar.mm
```diff
@@ -0,0 +1,327 @@
...
+                apiClient = [[MMEAPIClient alloc] initWithAccessToken:@"access-token"
+                                                        userAgentBase:@"user-agent-base"
+                                                       hostSDKVersion:@"host-sdk-1"];
+            });
+            
+            it(@"should use test server", ^{
+                apiClient.sessionWrapper.usesTestServer should be_truthy;
+            });
+            
+            it(@"should equal the BaseURL constant", ^{
```
boundsj: typo: this description should state that it should equal a URL made from the test URL string

MapboxMobileEventsTests/MMEAPIClientTests_cedar.mm
```diff
@@ -0,0 +1,327 @@
...
+                apiClient.baseURL should equal([NSURL URLWithString:testURLString]);
+            });
+            
+            afterEach(^{
+                stop_spying_on([NSUserDefaults standardUserDefaults]);
+            });
+        });
+        
+        context(@"when a bad URL is added to MMETelemetryTestServerURL", ^{
```
boundsj: Maybe say why it is bad? "when a non-secure URL is ..."

MapboxMobileEventsTests/MMEAPIClientTests_cedar.mm
```diff
@@ -0,0 +1,327 @@
...
+                [apiClient postEvents:@[event, eventTwo] completionHandler:nil];
+            });
+            
+            it(@"should equal gzip for HTTPHeader", ^{
+                sessionWrapperFake.request.allHTTPHeaderFields[MMEAPIClientHeaderFieldContentEncodingKey] should equal(@"gzip");
+            });
+            
+            it(@"should have compressed", ^{
+                NSData *data = (NSData *)sessionWrapperFake.request.HTTPBody;
+                data.length should be_less_than(uncompressedData.length);
```
boundsj: We could also compress the data in the test and check that the data instances are equal -- but what you have here seems like a reasonable way to do this here. 👍

MapboxMobileEventsTests/MMEAPIClientTests_cedar.mm
```diff
@@ -0,0 +1,327 @@
...
+            });
+        });
+        
+        context(@"when posting three events", ^{
```
boundsj: We could probably delete this since we already have coverage for the compressed case.

MapboxMobileEventsTests/MMEAPIClientTests_cedar.mm
```diff
@@ -0,0 +1,327 @@
...
+                        [eventAttributes addObject:event.attributes];
+                    }
+                }];
+                
+                uncompressedData = [NSJSONSerialization dataWithJSONObject:eventAttributes options:0 error:nil];
+                
+                [apiClient postEvents:@[event, eventTwo] completionHandler:nil];
+            });
+            
+            it(@"should equal gzip for HTTPHeader", ^{
```
boundsj: Maybe say "should use gzip for content encoding"

MapboxMobileEventsTests/MMEAPIClientTests_cedar.mm
```diff
@@ -0,0 +1,327 @@
...
+                sessionWrapperFake.request.allHTTPHeaderFields[MMEAPIClientHeaderFieldContentEncodingKey] should equal(@"gzip");
+            });
+            
+            it(@"should have compressed", ^{
```
boundsj: Maybe say "should compress the data"

boundsj (changes requested):
This is really close! 

Let's pair on the `postEvent:completionHandler:` section. I think I see a few ways we can express those tests more clearly.

boundsj (approved):

boundsj:
Part of https://github.com/mapbox/mapbox-events-ios/issues/12

PR #20, opened 2017-10-26
Remove warnings

rclee:
Also addresses potential error with accessing allowsBackgroundLocationUpdates on iOS devices running < 9.0

### DIFF
--- a/MapboxMobileEvents/MMEEventsManager.m
+++ b/MapboxMobileEvents/MMEEventsManager.m
@@ -72,7 +72,7 @@
     
     [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(pauseOrResumeMetricsCollectionIfRequired) name:UIApplicationDidEnterBackgroundNotification object:nil];
     [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(pauseOrResumeMetricsCollectionIfRequired) name:UIApplicationDidBecomeActiveNotification object:nil];
-    if (&NSProcessInfoPowerStateDidChangeNotification != NULL) {
+    if (@available(iOS 9.0, *)) {
         [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(pauseOrResumeMetricsCollectionIfRequired) name:NSProcessInfoPowerStateDidChangeNotification object:nil];
     }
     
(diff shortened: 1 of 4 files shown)


MapboxMobileEvents/MMEEventsManager.m
```diff
@@ -72,7 +72,7 @@ - (void)initializeWithAccessToken:(NSString *)accessToken userAgentBase:(NSStrin
     
     [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(pauseOrResumeMetricsCollectionIfRequired) name:UIApplicationDidEnterBackgroundNotification object:nil];
     [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(pauseOrResumeMetricsCollectionIfRequired) name:UIApplicationDidBecomeActiveNotification object:nil];
-    if (&NSProcessInfoPowerStateDidChangeNotification != NULL) {
+    if (@available(iOS 9.0, *)) {
```
boundsj: Unfortunately, I don't think we can use the @available feature since that is only supported since Xcode 9. This project needs to support development with prior versions of Xcode where @available did not yet exist for Objective-C.

boundsj: Also, this check needs to be a runtime check to help with the case of running on an older version of iOS even if the library was compiled with a newer iOS SDK.

boundsj (changes requested):
Looking good. Unfortunately I think we need to use the `respondsToSelector:` to guard against calls to `setAllowsBackgroundLocationUpdates` on older iOS versions.

MapboxMobileEvents/MMELocationManager.m
```diff
@@ -95,7 +96,7 @@ - (void)startLocationServices {
         if (authorizedAlways && self.hostAppHasBackgroundCapability) {
             [self.locationManager startMonitoringSignificantLocationChanges];
             [self startBackgroundTimeoutTimer];
-            if (@available(iOS 9.0, *)) {
+            if ([self respondsToSelector:@selector(allowsBackgroundLocationUpdates)]) {
```
boundsj: Great! Although, I think we still need to ignore the warning here, right?

rclee: Ah, you're right. Missed that--thanks!

boundsj (approved):
👍

PR #22, opened 2017-11-07
extend event logging for debugging purposes

rclee:
-adds new method to MMEEventLogger for writing events to a local file
-writes any event passing through sendTurnstile or pushEvent methods in MMEEventsManager

### DIFF
--- a/MapboxMobileEvents/MMEEvent.m
+++ b/MapboxMobileEvents/MMEEvent.m
@@ -5,20 +5,6 @@
 
 @implementation MMEEvent
 
-+ (instancetype)pushEventWithAttributes:(NSDictionary *)attributes {
-    MMEEvent *pushEvent = [[MMEEvent alloc] init];
-    pushEvent.name = MMEEventTypePush;
-    pushEvent.attributes = attributes;
-    return pushEvent;
-}
-
-+ (instancetype)flushEventWithAttributes:(NSDictionary *)attributes {
-    MMEEvent *flushEvent = [[MMEEvent alloc] init];
-    flushEvent.name = MMEEventTypeFlush;
-    flushEvent.attributes = attributes;
-    return flushEvent;
-}
-
 + (instancetype)turnstileEventWithAttributes:(NSDictionary *)attributes {
     MMEEvent *turnstileEvent = [[MMEEvent alloc] init];
     turnstileEvent.name = MMEEventTypeAppUserTurnstile;
--- a/MapboxMobileEvents/MMEEvent.m
+++ b/MapboxMobileEvents/MMEEvent.m
@@ -5,6 +5,20 @@
 
 @implementation MMEEvent
 
++ (instancetype)pushEventWithAttributes:(NSDictionary *)attributes {
+    MMEEvent *pushEvent = [[MMEEvent alloc] init];
+    pushEvent.name = MMEEventTypePush;
+    pushEvent.attributes = attributes;
+    return pushEvent;
+}
+
++ (instancetype)flushEventWithAttributes:(NSDictionary *)attributes {
+    MMEEvent *flushEvent = [[MMEEvent alloc] init];
+    flushEvent.name = MMEEventTypeFlush;
+    flushEvent.attributes = attributes;
+    return flushEvent;
+}
+
 + (instancetype)turnstileEventWithAttributes:(NSDictionary *)attributes {
     MMEEvent *turnstileEvent = [[MMEEvent alloc] init];
     turnstileEvent.name = MMEEventTypeAppUserTurnstile;
--- a/MapboxMobileEvents/MMEEventLogger.h
+++ b/MapboxMobileEvents/MMEEventLogger.h
@@ -6,9 +6,11 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface MMEEventLogger : NSObject
 
-+ (BOOL)isEnabled;
-+ (void)setEnabled:(BOOL)enabled;
-+ (void)logEvent:(MMEEvent *)event;
+@property (nonatomic, getter=isEnabled) BOOL enabled;
+
++ (instancetype)sharedLogger;
+
+- (void)logEvent:(MMEEvent *)event;
 
 - (void)writeEventToLocalDebugLog:(MMEEvent *)event;
 
--- a/MapboxMobileEvents/MMEEventLogger.h
+++ b/MapboxMobileEvents/MMEEventLogger.h
@@ -12,8 +12,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (void)logEvent:(MMEEvent *)event;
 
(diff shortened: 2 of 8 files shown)


MapboxMobileEvents/MMEEventLogger.h
```diff
@@ -10,6 +10,8 @@ NS_ASSUME_NONNULL_BEGIN
 + (void)setEnabled:(BOOL)enabled;
```
boundsj: We could probably make `enabled` a property.

boundsj: I forgot that this class has so far only used class methods to invoke logging (e.g. `[MMEEventLogger logEvent:debugEvent]`). This PR introduces the notion of using an instance of `MMEEventLogger` in `MMEEventsManager`. 

I think we should change the methods to make them instance based and add a singleton interface so that we can use a `[MMEEventLogger sharedLogger]` instance everywhere we want log messages to go to the same file with the same date. We can still create 1-off instances for other kinds of logging we may eventually want to do in different files.

MapboxMobileEvents/MMEEventLogger.m
```diff
@@ -18,4 +26,44 @@ + (void)setEnabled:(BOOL)enabled {
     _enabled = enabled;
 }
 
+- (void)writeEventToLocalDebugLog:(MMEEvent *)event {
```
boundsj: I think we can go ahead and combine this new method and the old `logEvent`. I like the name `logEvent` because it is easier to say. We can just assume that if something gets logged it is written to the console and to disk.

boundsj (commented):
Looking good! Please have a look at the comments wrt rationalizing how we use this class as a shared instance vs one off instances.

MapboxMobileEvents/MMEEventLogger.m
rclee: I definitely like this approach better. Initially I had similar thoughts with combining the two methods. My only concern was potentially having too many logs written locally and that a majority of the debug logs will have the same ``MMEEventKeyLocalDebug`` key.

It seems like we'll still need to have other keys for at least ``flush`` and ``push`` and that would likely result in any ``MMEEventKeyLocalDebug`` key/value pairs being unused; unless that information is interesting in some way...

boundsj: Let's go ahead and default to writing "too much" for now as a first pass. We can eventually introduce the notion of log levels to filter what actually gets printed and written.

MapboxMobileEvents/MMEEvent.m
```diff
@@ -5,6 +5,20 @@
 
 @implementation MMEEvent
 
++ (instancetype)pushEventWithAttributes:(NSDictionary *)attributes {
```
boundsj: Instead of making an API that would need to expand for new, arbitrary things that could all be categorized as "debug" events, let's reuse the existing `debugEventWithAttributes` API and plan on using the attributes for distinction.

I'm saying this also because I think of `push` and `flush` as actions that we want to track in a debug log but not as actual events that we send to the server, if that makes sense.

MapboxMobileEvents/MMEEventsManager.m
```diff
@@ -157,6 +157,12 @@ - (void)flush {
         return;
     }
     
+    NSDictionary *flushEventAttributes = @{MMEEventKeyEvent: MMEEventTypeFlush,
+                                          MMEEventKeyCreated: [self.dateWrapper formattedDateStringForDate:[self.dateWrapper date]]};
+    MMEEvent *flushEvent = [MMEEvent pushEventWithAttributes:flushEventAttributes];
```
boundsj: With the changes in this PR so far, I think this should actually be `flushEventWithAttributes:`. But let's make the change to use the unified `debugEventWithAttributes`. As a debug event, the event name will be `MMEEventTypeLocalDebug` (that is done for us by `debugEventWithAttributes:`) and we could introduce a "debug event type" that we put in the attributes. In this case, that could be a string constant called  `MMEDebugEventTypeFlush` that equals "flush".

MapboxMobileEvents/MMEEventsManager.m
```diff
@@ -375,7 +389,7 @@ - (void)pushDebugEventWithAttributes:(MMEMapboxEventAttributes *)attributes {
     [combinedAttributes setObject:[self.dateWrapper formattedDateStringForDate:[self.dateWrapper date]] forKey:@"created"];
     [combinedAttributes setObject:self.uniqueIdentifer.rollingInstanceIdentifer forKey:@"instance"];
     MMEEvent *debugEvent = [MMEEvent debugEventWithAttributes:combinedAttributes];
-    [MMEEventLogger logEvent:debugEvent];
+    [MMEEventLogger.sharedLogger logEvent:debugEvent];
```
boundsj, line 406: For all of the existing invocations of `pushDebugEventWithAttributes:` we should make sure their attributes have a `MMEDebugEventType` and `MMEDebugEventType{foo}` value that makes sense for the debug log entries they represent.

boundsj (changes requested):
Thanks for making the changes to allow for using the logger as a singleton. This helped illustrate for me some other changes I think we should make.

MapboxMobileEvents/MMEEventLogger.h
```diff
@@ -6,9 +6,13 @@ NS_ASSUME_NONNULL_BEGIN
...
-+ (BOOL)isEnabled;
-+ (void)setEnabled:(BOOL)enabled;
-+ (void)logEvent:(MMEEvent *)event;
+@property (nonatomic, getter=isEnabled) BOOL enabled;
+
++ (instancetype)sharedLogger;
+
+- (void)logEvent:(MMEEvent *)event;
+
+- (void)writeEventToLocalDebugLog:(MMEEvent *)event;
```
boundsj: I guess we could remove this in the header file and make it an implementation detail.

MapboxMobileEvents/MMEEventsManager.m
```diff
@@ -273,19 +290,21 @@ - (void)createAndPushEventBasedOnName:(NSString *)name attributes:(NSDictionary
     }
     
     if (event) {
-        [self pushDebugEventWithAttributes:@{MMEEventKeyLocalDebugDescription: [NSString stringWithFormat:@"Pushing event: %@", event]}];
+        [self pushDebugEventWithAttributes:@{MMEDebugEventType: MMEDebugEventTypePush,
```
boundsj, line 293: This is fine for now but for MMEDebugEventTypePush and MMEDebugEventTypeTurnstile similarly, we should probably make event sub-types to make it easier to visualize them when we build something to parse these logs. Otherwise MMEDebugEventTypePush success and failure case will be hard to discern unless we parse the MMEEventKeyLocalDebugDescription

boundsj (dismissed):

PR #26, opened 2017-12-01
Log reporter

rclee:
This adds a method to generate a view controller with a WKWebView to display event profiling information for debugging purposes.


MapboxMobileEvents/MMEEventLogReportViewController.m
```diff
@@ -0,0 +1,53 @@
...
+    configuration.userContentController = controller;
+    self.webView = [[WKWebView alloc] initWithFrame:CGRectMake(0, view.frame.size.height, self.view.frame.size.width, self.view.frame.size.height - view.frame.size.height) configuration:configuration];
+    self.webView.UIDelegate = self;
+    [self.view addSubview:self.webView];
+    
+    _spinner = [[UIActivityIndicatorView alloc] initWithActivityIndicatorStyle:UIActivityIndicatorViewStyleGray];
+    [_spinner setFrame:CGRectMake(self.view.frame.size.width/2 - _spinner.frame.size.width/2, self.view.frame.size.height/2, _spinner.frame.size.width, _spinner.frame.size.height)];
+    [self.view addSubview:_spinner];
+    
+    dispatch_async(dispatch_get_main_queue(), ^{
```
boundsj: What is the purpose of marshaling to the main thread here? `viewDidLoad` would never be called in any other thread context, right?

boundsj (changes requested):
Thanks! This feature will really help profile performance of this library.

In addition to the questions I asked in-line, I was wondering how the `readAndDisplayLogFileFromDate:viewController` method is intended to be invoked. I don't think `MMEEventLogger` is or ever should be public so it is not something a host SDK or application can use. I think that means we would need to add an API to `MMEEventsManager` and the host SDK/app would need to know that API exists and call it and it, in turn, would use `MMEEventLogger`. 

Injecting this profile/HTML view is going to be a bit awkward no matter what but I wonder about an internal, undocumented API that automatically finds the current, topmost view controller and presents using whatever that is. Alternatively, I suppose we could design an API that vends a web view and the host SDK/app can be responsible for adding that web view to the hierarchy.

rclee:
Great feedback! I'll look into adding these changes tonight. I figured the `MMEEventLogger` should probably remain private but since it was just for debugging purposes that we might be able to get away with it. In hindsight the design you described makes more sense. 

I'll look into implementing the API for finding the topmost view controller, that is certainly cleaner than the current implementation. Thanks!

MapboxMobileEvents/MMEEventLogReportViewController.m
rclee: Yes, you're right--just a oversight on my part. Thanks for catching that.

MapboxMobileEvents/MMEConstants.m
```diff
@@ -71,6 +71,8 @@
 
 NSString * const MMEEventSource = @"mapbox";
 
+NSString * const MMELoggerHTML = @"<html><head><script type='text/javascript' src='https://www.gstatic.com/charts/loader.js'></script><script type='text/javascript'>google.charts.load('current', {'packages':['timeline']});var dataString;function addData(data) {dataString = data}window.webkit.messageHandlers.data.postMessage('data');google.charts.setOnLoadCallback(drawChart); function drawChart() {var dataTable = new google.visualization.DataTable({cols: [{id: 'eventType', label: 'Event Type', type: 'string'},{id: 'start', label: 'Event Start Time', type: 'datetime'},{id: 'end', label: 'Event End Time', type: 'datetime'}],rows: dataString});var options = {'title':'Telemetry Log Data','width':1024,'height':400,'timeline': { groupByRowLabel: true }};var chart = new google.visualization.Timeline(document.getElementById('chart_div'));google.visualization.events.addListener(chart, 'ready', afterDraw);chart.draw(dataTable, options);}function afterDraw() {window.webkit.messageHandlers.complete.postMessage('complete');}</script></head><body><div id='chart_div'></div></body></html>";
```
boundsj, line 74: Thanks for putting this back and sorry for waffling from [my original comment](https://github.com/mapbox/mapbox-events-ios/pull/26/files/9956e4b2cb1e5fab798568b2ed2280c87e32bba9#r154723787). No need to change anything here but, in the future to make this more readable and easier to update, we could code generate this string from an html file that is in this project but not shipped as part of this library.

boundsj (approved):
Looks good to me!

PR #29, opened 2017-12-19
Fix bug with log file creation

rclee:
Also a light refactor on how we create a startOfTomorrow date, moving that function to the MMENSDateWrapper class

### DIFF
--- a/MapboxMobileEvents/MMEEventLogger.m
+++ b/MapboxMobileEvents/MMEEventLogger.m
@@ -8,10 +8,10 @@
 @interface MMEEventLogger()
 
 @property (nonatomic, copy) NSString *dateForDebugLogFile;
-@property (nonatomic) NSFileManager *fileManager;
 @property (nonatomic) dispatch_queue_t debugLogSerialQueue;
 @property (nonatomic) MMENSDateWrapper *dateWrapper;
 @property (nonatomic) NSDate *nextLogFileDate;
+@property (nonatomic) NSDateFormatter *dateFormatter;
 
 @end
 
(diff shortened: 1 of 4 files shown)


boundsj (changes requested):
Great fix and thanks for doing this! 

I think we could refactor a bit more to do the following:

* Add an init method to this class that:
  * allocates and initializes an `NSDateFormatter` for a new `dateFormatter` internal property
  * allocates and initializes a `MMENSDateWrapper` for the `dateWrapper` internal property
  * preemptively sets nextLogFileDate to the startOfTomorrow
  * preemptively sets `dateForDebugLogFile` using `newDateForLogFile`

then

* Replace usages of `newDateForLogFile` with `self.dateForDebugLogFile = [dateFormatter stringFromDate:[self.dateWrapper date]]` and delete `newDateForLogFile` method. Starting an objc method name with the word `new` is [not advisable](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/MemoryMgmt/Articles/mmRules.html) and that method would not be doing much anyway if the `NSDateFormatter` is created when the class is initialized.

This will allow us to get rid of most of the checks that are currently part of `writeEventToLocalDebugLog` and move them to an initializer so that setup does not need to be thought about more than once (at runtime). The only check remaining will be the one that checks if  it is currently after the `nextLogFileDate` so that the `dateForDebugLogFile` and `nextLogFileDate` can be updated if required.

To make that remaining check easier to read, you could consider wrapping the `[[self.dateWrapper date] timeIntervalSinceDate:self.nextLogFileDate] > 0` comparison in a new, internal BOOL property (e.g. `isTimeToUseANewLogFile`) so that check is bit more semantic.

rclee:
Great insight, I've added the changes you've detailed! Thanks 🙇

MapboxMobileEvents/MMEEventLogger.m
```diff
@@ -35,16 +51,18 @@ - (void)logEvent:(MMEEvent *)event {
 
 #pragma mark - Write to Local File
 
+- (BOOL)isTimeForNewLogFile {
```
boundsj, line 55: Another way of doing something like this is to make an internal property like `@property (nonatomic, readonly, getter=isTimeForNewLogFile) BOOL timeForNewLogFile`. The implementation you already have would implement the getter described in the property declaration.

boundsj (dismissed):
Thanks! 👍

MapboxMobileEvents/MMEEventLogger.m
rclee: Thanks for pointing this out. I've added the internal property, it should help it to be a bit more readable and match the style of the project.

PR #31, opened 2018-01-10
Logger improvements

rclee:
Change to the HTML to show a more detailed tooltip. Webview timeline is color coded in groups based on the instance. 

We will receive debug events for success or failure when attempting a post request.


boundsj (approved):
👍

PR #34, opened 2018-02-07
Update release package script

boundsj:
This updates the package script so that it can:

* Update version numbers in all relevant files for a release
* Make a local commit with the updated version
* Git tag a release for a version
* Push the version commit / tag to GitHub

It also adds a command to package up a static build and a few utilities to revert local commits / tags if you change your mind or make a typo.

cc @rclee


rclee (approved):
Thanks so much, this should really help to streamline the process.

PR #35, opened 2018-02-09
Re-introduce visit collection and events

boundsj:
This also refactors the MMEEventsManagerDelegate to reflect that it (the eventsManager) is the source of the delegate callbacks and not its location manager.


1ec5 (approved):

vincethecoder (commented):
LGTM 👍

PR #36, opened 2018-02-20
Port Radius Profile Configuration

rclee:
- Creates new class MMEEventsService to manage the creation and distribution of MMEEventsConfiguration objects/configurations.
- Expands MMEEventsConfiguration to include a variable hibernation radius configuration and potentially additional configurations in the future. A user may access the variable radius by setting both "MMEEventsProfile" (with the appropriate key "VariableGeofence") and "MMEGeofenceRadius" with a desired radius (within a range we've set) in the applications plist. If "MMEEventsProfile" is set in plist without choosing a radius then a median radius will be the default. 
- Modifies MMELocationManager and MMEEventsManager to use MMEEventsService configuration
- Adds and repairs cedar tests

### DIFF
--- a/MapboxMobileEvents.podspec
+++ b/MapboxMobileEvents.podspec
@@ -3,7 +3,7 @@ Pod::Spec.new do |s|
   # ―――  Spec Metadata  ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――― #
 
   s.name = 'MapboxMobileEvents'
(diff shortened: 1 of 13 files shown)


boundsj (approved):
This looks good to me. Let's hold off on merging for now until we have confirmation of the need for this change and a spec for the upper bound (kHibernationRadiusWide). Thanks!

PR #43, opened 2018-05-29
Add configuration profile options

rclee:
- Creates new class MMEEventsService to manage the creation and distribution of MMEEventsConfiguration objects/configurations.
- Modifies MMELocationManager and MMEEventsManager to use MMEEventsService configuration
- Adds and repairs cedar tests
- Fix bug with geofence creation related to custom profile
- Update namespace dependencies symbols

### DIFF
--- a/MapboxMobileEvents/MMEEventsManager.m
+++ b/MapboxMobileEvents/MMEEventsManager.m
@@ -12,6 +12,7 @@
 #import "MMENSDateWrapper.h"
 #import "MMECategoryLoader.h"
 #import "CLLocation+MMEMobileEvents.h"
+#import "MMEEventsService.h"
 #import <CoreLocation/CoreLocation.h>
 
 @interface MMEEventsManager () <MMELocationManagerDelegate>
@@ -37,12 +38,12 @@
 + (instancetype)sharedManager {
     static MMEEventsManager *_sharedManager;
     static dispatch_once_t onceToken;
-
+    
     dispatch_once(&onceToken, ^{
         [MMECategoryLoader loadCategories];
         _sharedManager = [[MMEEventsManager alloc] init];
     });
-
+    
     return _sharedManager;
 }
 
@@ -54,7 +55,7 @@
         _accountType = 0;
         _eventQueue = [NSMutableArray array];
         _commonEventData = [[MMECommonEventData alloc] init];
-        _configuration = [MMEEventsConfiguration defaultEventsConfiguration];
+        _configuration = [MMEEventsService.sharedService configuration];
         _uniqueIdentifer = [[MMEUniqueIdentifier alloc] initWithTimeInterval:_configuration.instanceIdentifierRotationTimeInterval];
         _application = [[MMEUIApplicationWrapper alloc] init];
         _dateWrapper = [[MMENSDateWrapper alloc] init];
@@ -73,12 +74,9 @@
     [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(pauseOrResumeMetricsCollectionIfRequired) name:UIApplicationDidEnterBackgroundNotification object:nil];
     [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(pauseOrResumeMetricsCollectionIfRequired) name:UIApplicationDidBecomeActiveNotification object:nil];
     
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wunguarded-availability"
-    if (&NSProcessInfoPowerStateDidChangeNotification != NULL) {
+    if (@available(iOS 9.0, *)) {
         [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(pauseOrResumeMetricsCollectionIfRequired) name:NSProcessInfoPowerStateDidChangeNotification object:nil];
     }
-#pragma clang diagnostic pop
     
     self.paused = YES;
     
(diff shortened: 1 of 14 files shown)


MapboxMobileEvents/MMEEventsManager.m
```diff
@@ -54,7 +55,7 @@ - (instancetype)init {
         _accountType = 0;
         _eventQueue = [NSMutableArray array];
         _commonEventData = [[MMECommonEventData alloc] init];
-        _configuration = [MMEEventsConfiguration defaultEventsConfiguration];
+        _configuration = [MMEEventsService.sharedService configuration];
```
fabian-guerra: Let's keep the bracket convention  `[[MMEEventsService sharedService] configuration]`

MapboxMobileEvents/MMEEventsService.m
```diff
@@ -0,0 +1,36 @@
...
+
+- (instancetype) init {
+    self = [super init];
+    if (self) {
+        self.configuration = [self configurationFromKey:[[NSBundle mainBundle] objectForInfoDictionaryKey:kMMEEventsProfile]];
+    }
+    return self;
+}
+
++ (nullable instancetype)sharedService {
```
fabian-guerra: Same question as above.
> Shouldn't we expect a singleton to be non-null?

fabian-guerra (commented):
Overall looks good. Just minor changes. Have you tested energy consumption with this change?

rclee:
@fabian-guerra Yes, battery testing has been done. Results were mostly as expected with the increase in geofence radius.

MapboxMobileEvents/MMELocationManager.m
```diff
@@ -30,6 +33,7 @@ - (instancetype)init {
         _application = [[MMEUIApplicationWrapper alloc] init];
         NSArray *backgroundModes = [[NSBundle mainBundle] objectForInfoDictionaryKey:@"UIBackgroundModes"];
         _hostAppHasBackgroundCapability = [backgroundModes containsObject:@"location"];
+        _configuration = [MMEEventsService.sharedService configuration];
```
fabian-guerra: I missed this one `[MMEEventsService sharedService]` please make sure all calls to the `sharedService` follows the convention.

akitchen:
LGTM. What's up with CI? Can you fix up merge conflicts and re-push before final 👍  ?

MapboxMobileEvents/MMELocationManager.m
rclee: Right, sorry about that. Thanks for catching those.

fabian-guerra (dismissed):
Thank you. 👍🏼

MapboxMobileEventsTests/MMEEventsManagerTests.mm
```diff
@@ -831,14 +831,21 @@ @interface MMEEventsManager (Tests) <MMELocationManagerDelegate>
             
             it(@"enqueues the correct event", ^{
                 CLLocation *location = [[CLLocation alloc] initWithLatitude:visit.coordinate.latitude longitude:visit.coordinate.longitude];
-                NSDictionary *attributes = @{MMEEventKeyCreated: [eventsManager.dateWrapper formattedDateStringForDate:[location timestamp]],
+                NSDictionary *attributes = @{MMEEventKeyCreated: @"2018-05-30T21:53:08.720+0000",
```
fabian-guerra: Have you consider using an `NSDate` + `NSDateFormatter` instead of hardcoding a string date?

MapboxMobileEventsTests/MMEEventsManagerTests.mm
```diff
@@ -831,14 +831,21 @@ @interface MMEEventsManager (Tests) <MMELocationManagerDelegate>
...
                                              MMEEventHorizontalAccuracy: @(visit.horizontalAccuracy),
                                              MMEEventKeyArrivalDate: [eventsManager.dateWrapper formattedDateStringForDate:visit.arrivalDate],
                                              MMEEventKeyDepartureDate: [eventsManager.dateWrapper formattedDateStringForDate:visit.departureDate]};
                 MMEEvent *expectedVisitEvent = [MMEEvent visitEventWithAttributes:attributes];
-                eventsManager.eventQueue.firstObject should equal(expectedVisitEvent);
+                MMEEvent *enqueueEvent = eventsManager.eventQueue.firstObject;
+                
+                NSMutableDictionary *tempDict = [[NSMutableDictionary alloc] init];
+                [tempDict addEntriesFromDictionary:enqueueEvent.attributes];
+                [tempDict setObject:@"2018-05-30T21:53:08.720+0000" forKey:@"created"];
```
fabian-guerra: Same question as above.

MapboxMobileEventsTests/MMEEventsManagerTests.mm
rclee: Right, that was what I was using before and what was causing an issue. When the enqueue event is created with the `didVisit:` method, then a "created" time is added to the dictionary via NSDate. After that the `expectedVisitEvent`'s "created" attribute was added at nearly the same time (via NSDate) but would sometimes be off by a second and cause the test to fail.

fabian-guerra: I was thinking more into a 'static' date you could use on both cases:
```
NSDictionary *attributes = @{MMEEventKeyCreated: [eventsManager.dateWrapper formattedDateStringForDate:[location timestamp]],
```
And
```
[tempDict setObject:[eventsManager.dateWrapper formattedDateStringForDate:[location timestamp]] forKey:@"created"];
```

`[location timestamp]` would be the same in both cases right?

rclee: Yes, you're right...I wasn't putting two and two together there. Thanks!

fabian-guerra (approved):
👍🏼

PR #51, opened 2018-07-03
Add useragent override for RN

rclee:


MapboxMobileEvents/MMEAPIClient.m
```diff
@@ -100,6 +100,11 @@ - (NSURLRequest *)requestForEvents:(NSArray *)events {
 }
 
 - (void)setupUserAgent {
+    if ([self.applicationBundle objectForInfoDictionaryKey:@"MMEMapboxUserAgentBase"] && [self.applicationBundle objectForInfoDictionaryKey:@"MMEMapboxHostSDKVersion"]) {
```
nitaliano: @rclee should the base and SDK version be split into separate conditionals? On Android I won't be overriding the SDK version so it stays as the native version in the meantime.

julianrex:
Are these overrides just for the telemetry lib? Would the user's expectation be that this would also override the user agent for requests from core?

rclee:
@julianrex Yes, the expectation is to override core to track metrics coming from React Native.

nitaliano (approved):
👍

Issue #50, opened 2018-07-02
MapBox crashes at [MGL_MMELocationManager startBackgroundTimeoutTimer] (MMELocationManager.m:158)


ghost:
We have a lot of crashes with this report.

Maybe I should mention that 2 of 3 apps work fine with the mapbox release 4.1. The app which makes problems has this property `MGLMapboxMetricsEnabledSettingShownInApp = true` 

> *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[__NSCFString setBackgroundLocationServiceTimeoutAllowedDate:]: unrecognized selector sent to instance 0x102c52820'

```
Code Type:       ARM-64
Parent Process:  ??? [1]

Date/Time:       2018-06-29T12:29:02Z
Launch Time:     2018-06-29T12:28:47Z
OS Version:      iPhone OS 11.2.6 (15D100)
Report Version:  104

Exception Type:  SIGSEGV
Exception Codes: SEGV_ACCERR at 0x4c98abeb8
Crashed Thread:  0

Application Specific Information:
objc_msgSend() selector name: setBackgroundLocationServiceTimeoutAllowedDate:

Thread 0 Crashed:
0   libobjc.A.dylib                      0x0000000183aa8430 objc_msgSend + 16
1   Mapbox                               0x0000000101adb680 -[MGL_MMELocationManager startBackgroundTimeoutTimer] (MMELocationManager.m:158)
2   Mapbox                               0x0000000101adb8b8 -[MGL_MMELocationManager locationManager:didUpdateLocations:] (MMELocationManager.m:183)
3   CoreLocation                         0x000000018ae497bc CLClientRetrieveData + 77408
4   CoreLocation                         0x000000018ae4901c CLClientRetrieveData + 75456
5   CoreLocation                         0x000000018ae316b4 CLClientInvalidate + 1000
6   CoreFoundation                       0x00000001847f3590 __CFRUNLOOP_IS_CALLING_OUT_TO_A_BLOCK__ + 16
7   CoreFoundation                       0x00000001847f2e60 __CFRunLoopDoBlocks + 284
8   CoreFoundation                       0x00000001847f10c8 __CFRunLoopRun + 2432
9   CoreFoundation                       0x0000000184710c58 CFRunLoopRunSpecific + 432
10  GraphicsServices                     0x00000001865bcf84 GSEventRunModal + 96
11  UIKit                                0x000000018de695c4 UIApplicationMain + 232
[...]

PR #53, opened 2018-07-31
Refactor timer code into separate utility class

julianrex:
This PR moves the timer code out of `MMELocationManager` to avoid circular references.
It also ensures that the location manager's `delegate` property is set to nil to avoid a crash. (`CLLocationManager`'s `delegate` property is declared as `assign` so needs to be manually `nil`'d.)

We may want to move this secondary bug fix into a separate PR, so we merge faster, but both of these changes are related to circular references so seemed appropriate.

@rclee I wanted to get this PR in, even though we're still discussing the related telemetry crash at https://github.com/mapbox/mapbox-events-ios/issues/50. 

Update: ~Please note,  there is a similar change we need to make in https://github.com/mapbox/mapbox-gl-native, both of these bugs could account for the above crash. (I'll update this PR with that change when I create it.)~

Update 2: The introduction of a custom location manager (via the `MGLCLLocationManager` protocol in https://github.com/mapbox/mapbox-gl-native/pull/12013) does away for the related change I mention above).

### DIFF
--- a/MapboxMobileEvents/MMEBackgroundLocationServiceTimeoutHandler.h
+++ b/MapboxMobileEvents/MMEBackgroundLocationServiceTimeoutHandler.h
@@ -3,7 +3,7 @@
 @class MMEBackgroundLocationServiceTimeoutHandler;
 @protocol MMEUIApplicationWrapper;
 
-@protocol MMEBackgroundLocationServiceTimeoutHandlerDelegate
+@protocol MMEBackgroundLocationServiceTimeoutDelegate
 - (BOOL)timeoutHandlerShouldCheckForTimeout:(MMEBackgroundLocationServiceTimeoutHandler *)handler;
 - (void)timeoutHandlerDidTimeout:(MMEBackgroundLocationServiceTimeoutHandler *)handler;
 - (void)timeoutHandlerBackgroundTaskDidExpire:(MMEBackgroundLocationServiceTimeoutHandler *)handler;
@@ -12,7 +12,7 @@
 
 @interface MMEBackgroundLocationServiceTimeoutHandler: NSObject
 
-@property (nonatomic, weak) id<MMEBackgroundLocationServiceTimeoutHandlerDelegate> delegate;
+@property (nonatomic, weak) id<MMEBackgroundLocationServiceTimeoutDelegate> delegate;
 @property (nonatomic, readonly) NSTimer *timer;
 - (instancetype)initWithApplication:(id<MMEUIApplicationWrapper>)application;
 - (void)startTimer;
--- a/MapboxMobileEvents/MMEBackgroundLocationServiceTimeoutHandler.h
+++ b/MapboxMobileEvents/MMEBackgroundLocationServiceTimeoutHandler.h
@@ -1,0 +1,18 @@
+#import <Foundation/Foundation.h>
+
+@class MMEBackgroundLocationServiceTimeoutHandler;
+@protocol MMEUIApplicationWrapper;
+
+@protocol MMEBackgroundLocationServiceTimeoutHandlerDelegate
+- (BOOL)timeoutHandlerShouldCheckForTimeout:(MMEBackgroundLocationServiceTimeoutHandler *)handler;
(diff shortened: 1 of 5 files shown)


MapboxMobileEvents/MMEBackgroundLocationServiceTimeoutHandler.m
```diff
@@ -0,0 +1,85 @@
...
+    id<MMEBackgroundLocationServiceTimeoutHandlerDelegate> delegate = self.delegate;
+
+    if (!delegate) {
+        dispatch_async(dispatch_get_main_queue(), ^{
+            [self stopTimer];
+        });
+        return;
+    }
+
+    if (![delegate timeoutHandlerShouldCheckForTimeout:self]) {
```
fabian-guerra, line 38: While this is an internal application, would it be good if we check first if the delegate already implements the method?

MapboxMobileEvents/MMEBackgroundLocationServiceTimeoutHandler.m
```diff
@@ -0,0 +1,85 @@
...
+        return;
+    }
+
+    __weak __typeof__(self) weakself = self;
+    NSAssert(self.backgroundTaskId == UIBackgroundTaskInvalid, @"Background task Id should be invalid");
+    self.backgroundTaskId = [self.application beginBackgroundTaskWithExpirationHandler:^{
+        __typeof__(self) strongSelf = weakself;
+
+        [strongSelf stopTimer];
+        [strongSelf.delegate timeoutHandlerBackgroundTaskDidExpire:strongSelf];
```
fabian-guerra, line 69: Same comment as above.

MapboxMobileEvents/MMELocationManager.m
```diff
@@ -1,39 +1,44 @@
...
+@property (nonatomic) MMEBackgroundLocationServiceTimeoutHandler *backgroundLocationServiceTimeoutTimerWrapper;
 @property (nonatomic) BOOL hostAppHasBackgroundCapability;
 @property (nonatomic) MMEEventsConfiguration *configuration;
 
 @end
 
 @implementation MMELocationManager
 
+- (void)dealloc {
+    _locationManager.delegate = nil;
```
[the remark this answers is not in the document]
julianrex: This is a great question. I had all the `stop...` variants in here, but since the location manager is being deallocated (and its delegate is going too) I'm assuming* that there's no need to call these. It's not clear from the documentation, but I think it's a fair assumption.

Do you know?

*uh oh.

MapboxMobileEvents/MMEBackgroundLocationServiceTimeoutHandler.m
julianrex: These aren't `optional` protocol methods, so there's no need to check to see if the delegate already implements it. The compiler will warn if the client hasn't implemented them.

Since these are really internal methods, I didn't see a need to make them optional.

MapboxMobileEvents/MMEBackgroundLocationServiceTimeoutHandler.h
```diff
@@ -0,0 +1,21 @@
+#import <Foundation/Foundation.h>
+
+@class MMEBackgroundLocationServiceTimeoutHandler;
+@protocol MMEUIApplicationWrapper;
+
+@protocol MMEBackgroundLocationServiceTimeoutHandlerDelegate
```
julianrex: I hate this protocol name 😢 - any suggestions @rclee @fabian-guerra. Maybe replace `HandlerDelegate` with just one of `Protocol`, `Delegate`, `Handler`?

rclee: `MMEBackgroundLocationServiceTimeoutDelegate` works for me.

fabian-guerra (approved):
Please rebase and test the deallocation before merge.

julianrex:
Assigning to @rclee so that further tests can be run.

julianrex:
Closing, in preference of a `performSelector:withObject:afterDelay:` / `[NSObject cancelPreviousPerformRequestsWithTarget:selector:object:]` approach.

Please re-open if necessary.

Issue #54, opened 2018-08-01
Mapbox crashes randomly, normally in low gps areas or after hard closing app.


sillerjp:
```
0                                                                       testappiOS                     0x00000001039f6c10 mono_handle_native_crash + 224
2018-07-31 18:39:58.194 testappiOS[3659:1456484] critical:     1   testappiOS                     0x0000000103a0376c mono_sigsegv_signal_handler + 316
2018-07-31 18:39:58.194 testappiOS[3659:1456484] critical:     2   libsystem_platform.dylib            0x0000000183190b58 _sigtramp + 52
2018-07-31 18:39:58.194 testappiOS[3659:1456484] critical:     3   Mapbox                              0x0000000108757680 _Z45MBGLCustomGeometrySourceOptionsFromDictionaryP12NSDictionaryIP8NSStringP11objc_objectE + 180512
2018-07-31 18:39:58.194 testappiOS[3659:1456484] critical:     4   Mapbox                              0x00000001087578b8 _Z45MBGLCustomGeometrySourceOptionsFromDictionaryP12NSDictionaryIP8NSStringP11objc_objectE + 181080
2018-07-31 18:39:58.194 testappiOS[3659:1456484] critical:     5   CoreLocation                        0x0000000189d0e7b0 CLClientRetrieveData + 72260
2018-07-31 18:39:58.194 testappiOS[3659:1456484] critical:     6   CoreLocation                        0x0000000189d0e014 CLClientRetrieveData + 70312
2018-07-31 18:39:58.194 testappiOS[3659:1456484] critical:     7   CoreLocation                        0x0000000189cf79e8 CLClientInvalidate + 1240
2018-07-31 18:39:58.194 testappiOS[3659:1456484] critical:     8   CoreFoundation                      0x00000001835172bc <redacted> + 20
2018-07-31 18:39:58.194 testappiOS[3659:1456484] critical:     9   CoreFoundation                      0x0000000183516a7c <redacted> + 264
2018-07-31 18:39:58.194 testappiOS[3659:1456484] critical:     10  CoreFoundation                      0x00000001835147b0 <redacted> + 1224
2018-07-31 18:39:58.195 testappiOS[3659:1456484] critical:     11  CoreFoundation                      0x0000000183434da8 CFRunLoopRunSpecific + 552
2018-07-31 18:39:58.195 testappiOS[3659:1456484] critical:     12  GraphicsServices                    0x000000018541a020 GSEventRunModal + 100
[...]

PR #55, opened 2018-08-03
Set location manager's delegate to nil before deallocating

julianrex:
`CLLocationManager`'s `delegate` property is declared as `assign` - so needs to be nil'd whenever the instance containing the location manager is about to be deallocated.

This may partially address https://github.com/mapbox/mapbox-events-ios/issues/54 and https://github.com/mapbox/mapbox-events-ios/issues/50 - though it also requires a related change in `MGLMapView`.


rclee (approved):
Thanks @julianrex  👍

PR #56, opened 2018-08-03
Public key pinning 

rclee:
Updating public key pinning to use new hashes and configuration endpoint.


MapboxMobileEvents/MMEAPIClient.m
```diff
@@ -66,6 +74,34 @@ - (void)setBaseURL:(NSURL *)baseURL {
...
+- (NSError *)statusErrorFromRequest:(NSURLRequest *)request andHTTPResponse:(NSHTTPURLResponse *)httpResponse {
+    NSError *statusError = nil;
+    if (httpResponse.statusCode >= 400) {
+        NSString *descriptionFormat = @"The session data task failed. Original request was: %@";
+        NSString *reasonFormat = @"The status code was %ld";
+        NSString *description = [NSString stringWithFormat:descriptionFormat, request];
+        NSString *reason = [NSString stringWithFormat:reasonFormat, (long)httpResponse.statusCode];
+        NSDictionary *userInfo = @{NSLocalizedDescriptionKey: description,
+                                   NSLocalizedFailureReasonErrorKey: reason};
+        statusError = [NSError errorWithDomain:MMEErrorDomain code:1 userInfo:userInfo];
```
julianrex: I'd prefer to see an enum or similar for the error code here. It seems likely that we'd have other codes under the same `MMEErrorDomain`.

It might also be worth making it easier to access the HTTP status code from the info dictionary. As it stands you'd have to extract it from a format string. Consider adding the `NSHTTPURLResponse` to the dictionary.

MapboxMobileEvents/MMEConfigurationUpdater.h
```diff
@@ -0,0 +1,20 @@
+#import <Foundation/Foundation.h>
+#import "MMEEventsConfiguration.h"
+#import "MMEAPIClient.h"
+
+@protocol MMEConfigurationUpdaterDelegate <NSObject>
+
+- (void)configurationDidUpdate:(MMEEventsConfiguration *)configuration;
```
julianrex: Delegate methods (by convention) have a parameter for the sender. So (and I struggle with naming delegates), perhaps something like:
```
- (void)configurator:(MMEConfigurationUpdater *)updater didUpdate:(MMEEventsConfiguration *)configuration;
```

MapboxMobileEvents/MMEConfigurationUpdater.h
```diff
@@ -0,0 +1,20 @@
...
+
+@end
+
+@interface MMEConfigurationUpdater : NSObject
+
+- (instancetype)init __attribute__((unavailable("This method is not available")));
+- (instancetype)initWithTimeInterval:(NSTimeInterval)timeInterval;
```
julianrex: Perhaps use `NS_DESIGNATED_INITIALIZER` here?

MapboxMobileEvents/MMEConfigurationUpdater.m
```diff
@@ -0,0 +1,59 @@
+#import "MMEConfigurationUpdater.h"
+#import "MMEAPIClient.h"
+
+@interface MMEConfigurationUpdater ()
+
+@property (nonatomic) NSDate *configurationRotationDate;
+@property (nonatomic) MMEEventsConfiguration *configuration;
```
julianrex: This should be a `copy` property.

MapboxMobileEvents/MMEConfigurationUpdater.m
```diff
@@ -0,0 +1,59 @@
...
+
+- (void)updateConfigurationFromAPIClient:(MMEAPIClient *)apiClient {
+    if (self.configurationRotationDate && [[NSDate date] timeIntervalSinceDate:self.configurationRotationDate] >= 0) {
+        self.configuration = nil;
+    }
+    if (!self.configuration) {
+        [apiClient getConfigurationWithCompletionHandler:^(NSError * _Nullable error, NSData * _Nullable data) {
+            if (!error) {
+                self.configuration = [MMEEventsConfiguration configuration];
+                [self parseJSONFromData:data];
```
julianrex: Could we combine these two calls into one?  e.g.
```
+[MMEEventsConfiguration configurationFromData:]
```

It's confusing as it is, and `parseJSONFromData` only modifies the configuration.

MapboxMobileEvents/MMEEventsManager.m
```diff
@@ -202,6 +212,8 @@ - (void)flush {
 }
 
 - (void)sendTurnstileEvent {
+    [self.configurationUpdater updateConfigurationFromAPIClient:self.apiClient];
```
julianrex, line 234: Does the rest of `sendTurnstileEvent` (or any event for that matter) need to happen once the configuration is complete?

julianrex (commented):
Could this PR be broken into 2? 1 for the TrustKit changes and 1 for the configuration changes?

MapboxMobileEvents/MMEEventsManager.m
rclee: Negative. There aren't any plans currently for configurations to affect events directly.

MapboxMobileEvents/MMEAPIClient.m
```diff
@@ -66,6 +74,34 @@ - (void)setBaseURL:(NSURL *)baseURL {
...
+    NSString *path = [NSString stringWithFormat:@"%@?access_token=%@", MMEAPIClientEventsConfigPath, [self accessToken]];
+    
+    NSURL *url = [NSURL URLWithString:path relativeToURL:[NSURL URLWithString:MMEAPIClientBaseAPIURL]];
+    NSMutableURLRequest *request = [[NSMutableURLRequest alloc] initWithURL:url];
+    
+    [request setValue:self.userAgent forHTTPHeaderField:MMEAPIClientHeaderFieldUserAgentKey];
+    [request setValue:MMEAPIClientHeaderFieldContentTypeValue forHTTPHeaderField:MMEAPIClientHeaderFieldContentTypeKey];
+    [request setHTTPMethod:MMEAPIClientHTTPMethodGet];
+    
+    return [request copy];
```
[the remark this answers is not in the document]
rclee: I'm not sure honestly. I didn't add this line but it doesn't really seem necessary does it?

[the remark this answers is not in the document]
julianrex: Yeah, I don't think you need it here.

MapboxMobileEvents/MMEEventsManager.m
julianrex: 👍

MapboxMobileEvents/MMEAPIClient.m
```diff
@@ -54,18 +53,60 @@ - (void)postEvent:(MMEEvent *)event completionHandler:(nullable void (^)(NSError
...
+    
+    [self.sessionWrapper processRequest:request completionHandler:^(NSData * _Nullable data, NSURLResponse * _Nullable response, NSError * _Nullable error) {
+        if ([response isKindOfClass:[NSHTTPURLResponse class]]) {
+            NSHTTPURLResponse *httpResponse = (NSHTTPURLResponse *)response;
+            NSError *statusError = [self statusErrorFromRequest:request andHTTPResponse:httpResponse];
+            if (completionHandler) {
+                error = error ?: statusError;
+                completionHandler(error, data);
+            }
+        }
```
julianrex: We need to call the completion handler in all cases, so we need a `else` to this clause.

MapboxMobileEvents/MMEAPIClient.m
```diff
@@ -29,20 +33,15 @@ - (instancetype)initWithAccessToken:(NSString *)accessToken userAgentBase:(NSStr
...
 
+- (void)reconfigure:(MMEEventsConfiguration *)configuration {
+    [self.sessionWrapper reconfigure:configuration];
+}
+
 - (void)postEvents:(NSArray *)events completionHandler:(nullable void (^)(NSError * _Nullable error))completionHandler {
     NSURLRequest *request = [self requestForEvents:events];
     [self.sessionWrapper processRequest:request completionHandler:^(NSData * _Nullable data, NSURLResponse * _Nullable response, NSError * _Nullable error) {
-        NSError *statusError = nil;
         NSHTTPURLResponse *httpResponse = (NSHTTPURLResponse *)response;
```
julianrex: You have another cast here - so same kind of check probably needs to go here?

MapboxMobileEvents/MMETimerManager.m
```diff
@@ -19,6 +19,8 @@ - (instancetype)initWithTimeInterval:(NSTimeInterval)timeInterval target:(id)aTa
 
 - (void)start {
     [self cancel];
+    // NOTE: this has the side-effect of scheduling the timer on the current or main run loop.
+    // Once delayed startup is implemented, the scheduling of this timer may also need to be delayed
```
akitchen, line 23: Is there a plan to test this pattern in the field, based on this comment? We need to be careful about the memory management implications of using `NSTimer`. I like that it's encapsulated in a manager class which should make it easier to break any potential memory retain cycles, but nonetheless it's something we're going to want to put some additional thought into.

rclee: I think we could remove this comment. The delayed startup has been field tested to a fair extent at this point. Agreed on points of `NSTimer` - certainly something to keep an eye on. 👁

MapboxMobileEvents/MMEEventsManager.m
```diff
@@ -70,22 +76,28 @@ - (void)dealloc {
 
 - (void)initializeWithAccessToken:(NSString *)accessToken userAgentBase:(NSString *)userAgentBase hostSDKVersion:(NSString *)hostSDKVersion {
     self.apiClient = [[MMEAPIClient alloc] initWithAccessToken:accessToken userAgentBase:userAgentBase hostSDKVersion:hostSDKVersion];
+    self.configurationUpdater.delegate = self;
     
-    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(pauseOrResumeMetricsCollectionIfRequired) name:UIApplicationDidEnterBackgroundNotification object:nil];
-    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(pauseOrResumeMetricsCollectionIfRequired) name:UIApplicationDidBecomeActiveNotification object:nil];
+    void(^initialization)(void) = ^{
```
julianrex: I think we should do the `weakself`/`strongself` dance here, and then return early if `strongself` is `nil`.  My concern is the timer manager at the bottom of the block and any potential retain cycles.

MapboxMobileEvents/MMEEventsManager.m
```diff
@@ -454,6 +468,20 @@ - (void)displayLogFileFromDate:(NSDate *)logDate {
     [MMEEventLogger.sharedLogger readAndDisplayLogFileFromDate:logDate];
 }
 
+#pragma mark - MMEConfiguratorDelegate
+
+- (void)configurator:(id)updater didUpdate:(MMEEventsConfiguration *)configuration {
+    self.configuration = configuration;
+    [self.apiClient reconfigure:configuration];
+    [self.locationManager reconfigure:configuration];
```
julianrex: `reconfigure` is marked as `optional` - we should check to see if it responds to the selector (unless I'm looking at the wrong thing!)

MapboxMobileEvents/MMEAPIClient.m
```diff
@@ -181,6 +210,46 @@ - (NSURLRequest *)requestForEvents:(NSArray *)events {
...
+        [httpBody appendData:[[NSString stringWithFormat:@"--%@\r\n", boundary] dataUsingEncoding:NSUTF8StringEncoding]];
+        [httpBody appendData:[[NSString stringWithFormat:@"Content-Disposition: form-data; name=\"file\"; filename=\"%@\"\r\n", filename] dataUsingEncoding:NSUTF8StringEncoding]];
+        [httpBody appendData:[[NSString stringWithFormat:@"Content-Type: %@\r\n\r\n", mimetype] dataUsingEncoding:NSUTF8StringEncoding]];
+        [httpBody appendData:data];
+        [httpBody appendData:[@"\r\n\r\n" dataUsingEncoding:NSUTF8StringEncoding]];
+    }
+    
+    [httpBody appendData:[[NSString stringWithFormat:@"--%@--\r\n", boundary] dataUsingEncoding:NSUTF8StringEncoding]];
+    
+    NSLog(@"httpBody: %@",[[NSString alloc] initWithData:httpBody encoding:NSUTF8StringEncoding]);
```
julianrex: We should remove this (or potentially replace with the telemetry log?)

MapboxMobileEvents/MMEAPIClient.m
```diff
@@ -31,30 +36,39 @@ - (instancetype)initWithAccessToken:(NSString *)accessToken userAgentBase:(NSStr
...
-                                       NSLocalizedFailureReasonErrorKey: reason};
-            statusError = [NSError errorWithDomain:MMEErrorDomain code:1 userInfo:userInfo];
-        }
-        if (completionHandler) {
-            error = error ?: statusError;
-            completionHandler(error);
+        if ([response isKindOfClass:[NSHTTPURLResponse class]]) {
+            NSHTTPURLResponse *httpResponse = (NSHTTPURLResponse *)response;
+            statusError = [self statusErrorFromRequest:request andHTTPResponse:httpResponse];
+            if (completionHandler) {
```
julianrex: This `if` block can move below the parent `if`, and then you can remove the duplicate one below.

MapboxMobileEvents/MMEAPIClient.m
```diff
@@ -75,8 +89,26 @@ - (void)postMetadata:(NSArray *)metadata filepaths:(NSArray *)filepaths completi
...
-    [self postEvents:@[event] completionHandler:completionHandler];
+- (void)getConfigurationWithCompletionHandler:(nullable void (^)(NSError * _Nullable error, NSData * _Nullable data))completionHandler {
+    NSURLRequest *request = [self requestForConfiguration];
+    
+    [self.sessionWrapper processRequest:request completionHandler:^(NSData * _Nullable data, NSURLResponse * _Nullable response, NSError * _Nullable error) {
+        NSError *statusError = nil;
+        if ([response isKindOfClass:[NSHTTPURLResponse class]]) {
+            NSHTTPURLResponse *httpResponse = (NSHTTPURLResponse *)response;
+            statusError = [self statusErrorFromRequest:request andHTTPResponse:httpResponse];
+            if (completionHandler) {
```
julianrex: Ditto here.

MapboxMobileEvents/MMEAPIClient.m
julianrex: Noting that this has been addressed.

MapboxMobileEvents/MMEAPIClient.m
```diff
@@ -181,6 +213,49 @@ - (NSURLRequest *)requestForEvents:(NSArray *)events {
...
 
+- (NSString *)mimeTypeForPath:(NSString *)path {
+    CFStringRef extension = (__bridge CFStringRef)[path pathExtension];
+    CFStringRef UTI = UTTypeCreatePreferredIdentifierForTag(kUTTagClassFilenameExtension, extension, NULL);
+    if (UTI == NULL) {
+        return nil;
+    }
+    
+    NSString *mimetype = CFBridgingRelease(UTTypeCopyPreferredTagWithClass(UTI, kUTTagClassMIMEType));
+    if (mimetype == NULL) {
```
julianrex: This `if` block isn't needed, since you return `mimetype` anyway. If it did return here, you'd also have an unreleased `UTI`.

MapboxMobileEvents/MMEAPIClient.m
```diff
@@ -181,6 +213,49 @@ - (NSURLRequest *)requestForEvents:(NSArray *)events {
...
+    [httpBody appendData:[[NSString stringWithFormat:@"\r\n\r\n"] dataUsingEncoding:NSUTF8StringEncoding]];
+    
+    for (NSString *path in filePaths) {
+        NSString *filename  = [path lastPathComponent];
+        NSData   *data      = [NSData dataWithContentsOfFile:path];
+        NSString *mimetype  = [self mimeTypeForPath:path];
+        
+        [httpBody appendData:[[NSString stringWithFormat:@"--%@\r\n", boundary] dataUsingEncoding:NSUTF8StringEncoding]];
+        [httpBody appendData:[[NSString stringWithFormat:@"Content-Disposition: form-data; name=\"file\"; filename=\"%@\"\r\n", filename] dataUsingEncoding:NSUTF8StringEncoding]];
+        [httpBody appendData:[[NSString stringWithFormat:@"Content-Type: %@\r\n\r\n", mimetype] dataUsingEncoding:NSUTF8StringEncoding]];
```
julianrex, line 238: What happens here if `mimetype` is `nil`? I'd be tempted to wrap the appendData line with an `if (mimetype != nil)`.

MapboxMobileEvents/MMEEventsConfiguration.m
```diff
@@ -13,22 +19,62 @@ + (instancetype)defaultEventsConfiguration {
...
     configuration.locationManagerHibernationRadius = radius;
+    configuration.initializationDelay = delay;
     return configuration;
 }
 
++ (instancetype)configurationWithInfoDictionary:(NSDictionary *)infoDictionary {
+    NSString *profileName = infoDictionary[kMMEEventsProfile];
+    if ([profileName isEqualToString:kMMECustomProfile]) {
+        id customRadiusNumber = infoDictionary[kMMERadiusSize];
+        CLLocationDistance customRadius = [customRadiusNumber isKindOfClass:[NSNumber class]] ? [customRadiusNumber doubleValue] : 1200.0;
```
julianrex: Instead of `1200.0` let's use a named constant.

MapboxMobileEvents/MMEEventsConfiguration.m
```diff
@@ -13,22 +19,62 @@ + (instancetype)defaultEventsConfiguration {
...
+
+        id initializationDelayNumber = infoDictionary[kMMEStartupDelay];
+        NSTimeInterval initializationDelay = [initializationDelayNumber isKindOfClass:[NSNumber class]] ? [initializationDelayNumber doubleValue] : 10;
```
julianrex: Same here for the `10`

MapboxMobileEvents/MMEAPIClient.m
rclee: It's actually accepted server-side without a mimetype at the moment.

MapboxMobileEvents/MMEAPIClient.m
rclee: Good point

julianrex (approved):

PR #57, opened 2018-08-06
allow the vision namespace for vision sdk events.  

mick:
follows the same pattern as navigation events.

I've never commit code to this repo. Guidance on next steps, or @rclee or @zugaldia you can run with this from here, that would be appreciated.

cc @rclee @zugaldia  @deniskoronchik @l-r @willwhite


rclee (approved):
Looks good, thanks @mick 👍

PR #60, opened 2018-08-22
Trustkit configuration

rclee:
Adds new hashes and prepares for upcoming backend configurations.

### DIFF
--- a/MapboxMobileEvents/MMEHashProvider.h
+++ b/MapboxMobileEvents/MMEHashProvider.h
@@ -3,8 +3,8 @@
 
 @interface MMEHashProvider : NSObject
 
-@property (nonatomic) NSArray *cnHashes;
-@property (nonatomic) NSArray *comHashes;
+@property (nonatomic, readonly) NSArray *cnHashes;
+@property (nonatomic, readonly) NSArray *comHashes;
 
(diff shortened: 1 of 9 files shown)


MapboxMobileEvents/MMEHashProvider.h
```diff
@@ -0,0 +1,11 @@
+#import <Foundation/Foundation.h>
+#import "MMEEventsConfiguration.h"
+
+@interface MMEHashProvider : NSObject
+
+@property (nonatomic) NSArray *cnHashes;
+@property (nonatomic) NSArray *comHashes;
```
julianrex: Are these only supposed to be set by `updateHashesWithConfiguration `? If so, you might want to declare these as `readonly` in the header file, then redeclare as `readwrite` in the implementation file.

julianrex (dismissed):
One minor question, but looks good

MapboxMobileEvents/MMEHashProvider.h
rclee: Yes! Good call

rclee:
Closing: Merged back with https://github.com/mapbox/mapbox-events-ios/issues/56

PR #62, opened 2018-08-23
Call completion block on the main queue.

julianrex:
The assumption here is that `processRequest:completionHandler:` is being call on the main thread, and so the completion block should too. Is this a safe assumption? 

/cc @lloydsheng


rclee (dismissed):
I think we're good to go here.

julianrex:
Rebased

rclee (approved):

PR #68, opened 2018-09-13
Fix offline crash

rclee:
Crash that happens when there is no network connection and the response is `nil`. This PR changes error constructor methods to use `NSMutableDictionary` and `setValue:` to allow for `nil` values.

Fixes https://github.com/mapbox/mapbox-events-ios/issues/69

### DIFF
--- a/MapboxMobileEvents/MMEAPIClient.h
+++ b/MapboxMobileEvents/MMEAPIClient.h
@@ -20,8 +20,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (void)getConfigurationWithCompletionHandler:(nullable void (^)(NSError * _Nullable error, NSData * _Nullable data))completionHandler;
 - (void)reconfigure:(MMEEventsConfiguration *)configuration;
-- (NSError *)statusErrorFromRequest:(nonnull NSURLRequest *)request andHTTPResponse:(nonnull NSHTTPURLResponse *)httpResponse;
-- (NSError *)unexpectedResponseErrorfromRequest:(nonnull NSURLRequest *)request andResponse:(NSURLResponse *)response;
+- (NSError *)statusErrorFromRequest:(NSURLRequest *)request andHTTPResponse:(NSHTTPURLResponse *)httpResponse;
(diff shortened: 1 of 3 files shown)


MapboxMobileEvents/MMEAPIClient.m
```diff
@@ -116,6 +116,9 @@ - (void)setBaseURL:(NSURL *)baseURL {
 #pragma mark - Utilities
 
 - (NSError *)statusErrorFromRequest:(NSURLRequest *)request andHTTPResponse:(NSHTTPURLResponse *)httpResponse {
```
andrlee: Probably a stupid question, but can `(NSURLRequest *)request `param be `nil`? And what will happen if it is?

julianrex (dismissed):

MapboxMobileEvents/MMEAPIClient.m
rclee: Hasn't happened yet AFAIK but adding in some precautions won't hurt. 👍

MapboxMobileEvents/MMEAPIClient.h
```diff
@@ -20,6 +20,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (void)getConfigurationWithCompletionHandler:(nullable void (^)(NSError * _Nullable error, NSData * _Nullable data))completionHandler;
 - (void)reconfigure:(MMEEventsConfiguration *)configuration;
+- (NSError *)statusErrorFromRequest:(nonnull NSURLRequest *)request andHTTPResponse:(nonnull NSHTTPURLResponse *)httpResponse;
```
friedbunny: `nonnull` is implicit for all methods in this header because it uses the [`NS_ASSUME_NONNULL_BEGIN/END` macros](https://developer.apple.com/swift/blog/?id=25), so specifying it is redundant and unnecessary.

friedbunny:
Please add a description of the cause of the crash and what this PR does to fix it.

JThramer (approved):
:shipit:

PR #77, opened 2018-10-09
Carplay events modified to adhere to server schema.

vincethecoder:
- Updated the carplay event constants to match up with the expected events schema at the server side.

### DIFF
--- a/MapboxMobileEvents/MMENamespacedDependencies.h
+++ b/MapboxMobileEvents/MMENamespacedDependencies.h
@@ -425,6 +425,10 @@
 #define MMEEventTypeLocalDebug __NS_SYMBOL(MMEEventTypeLocalDebug)
 #endif
 
+#ifndef MMEEventTypeOfflineDownload
+#define MMEEventTypeOfflineDownload __NS_SYMBOL(MMEEventTypeOfflineDownload)
+#endif
+
 #ifndef MMEEventGestureSingleTap
 #define MMEEventGestureSingleTap __NS_SYMBOL(MMEEventGestureSingleTap)
 #endif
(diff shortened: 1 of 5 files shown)


rclee (approved):
👍

MapboxMobileEvents/MMENamespacedDependencies.h
```diff
@@ -425,6 +425,10 @@
 #define MMEEventTypeLocalDebug __NS_SYMBOL(MMEEventTypeLocalDebug)
 #endif
 
+#ifndef MMEEventTypeOfflineDownload
```
akitchen, line 428: What's this about? Is this related to CarPlay events at all?

MapboxMobileEventsTests/MMEEventsManagerTests.mm
```diff
@@ -829,20 +829,6 @@ @interface MMEEventsManager (Tests) <MMELocationManagerDelegate>
                 });
             });
             
-            context(@"when a carplay event is pushed", ^{
```
akitchen, line 832: Why delete this test? It seems it should be updated for the new event name?

MapboxMobileEvents/MMENamespacedDependencies.h
rclee: No, this is a separate event related to Maps SDK

MapboxMobileEventsTests/MMEEventsManagerTests.mm
rclee: It's no longer needed since the prefix check for `navigation` already exists

PR #76, opened 2018-10-03
Add mobile offline event

captainbarbosa:
Closes https://github.com/mapbox/mapbox-gl-native/issues/12751.

PR #80, opened 2018-10-11
Add offline download start/complete events

captainbarbosa:
Amends https://github.com/mapbox/mapbox-events-ios/pull/76/ by splitting out the mobile offline event into two events - one that will be called when an offline map download is started, and one when an offline map download has completed.

cc @osana @zugaldia


rclee (changes requested):
Run `make name-header` and update the branch then I think we're good 👍

rclee (approved):

PR #82, opened 2018-10-16
Telemetry metrics event

rclee:
Adds `MMEMetricsManager` singleton class to help manage counters across the library. Once metrics have been collected they should be sent every 24hr period UTC. 

This event is based off of this [ticket](https://github.com/mapbox/mobile-telemetry/issues/195) and the PR for [this issue](https://github.com/mapbox/mobile-telemetry/issues/228).

cc: @mapbox/mobile-telemetry


MapboxMobileEvents/MMEMetricsManager.h
```diff
@@ -0,0 +1,38 @@
...
+@property (nonatomic, readonly) int requests;
+@property (nonatomic, readonly) long totalDataTransfer;
+@property (nonatomic, readonly) long cellDataTransfer;
+@property (nonatomic, readonly) long wifiDataTransfer;
+@property (nonatomic, readonly) int appWakeups;
+@property (nonatomic, readonly) int eventCountFailed;
+@property (nonatomic, readonly) int eventCountTotal;
+@property (nonatomic, readonly) int eventCountMax;
+@property (nonatomic, readonly) int deviceTimeDrift;
+@property (nonatomic, readonly) float deviceLat;
```
andrlee: are we not cutting down the precision of lat/lon? should we stick to the typealias CLLocationDegrees which maps to double?

MapboxMobileEvents/MMEMetricsManager.h
```diff
@@ -0,0 +1,38 @@
...
+@interface MMEMetricsManager : NSObject
+
+@property (nonatomic, readonly) int requests;
+@property (nonatomic, readonly) long totalDataTransfer;
+@property (nonatomic, readonly) long cellDataTransfer;
+@property (nonatomic, readonly) long wifiDataTransfer;
+@property (nonatomic, readonly) int appWakeups;
+@property (nonatomic, readonly) int eventCountFailed;
+@property (nonatomic, readonly) int eventCountTotal;
+@property (nonatomic, readonly) int eventCountMax;
```
andrlee: should counts be all unsigned ints?

MapboxMobileEvents/MMEMetricsManager.m
```diff
@@ -0,0 +1,160 @@
...
+        for (MMEEvent *event in eventQueue) {
+            NSNumber *eventCount = [self.eventCountPerType objectForKey:event.name];
+            eventCount = [NSNumber numberWithInteger:[eventCount integerValue] + 1];
+            [self.eventCountPerType setObject:eventCount forKey:event.name];
+        }
+    }
+}
+
+- (void)metricsFromEvents:(nullable NSArray *)events error:(nullable NSError *)error {
+    if (error == nil) {
```
andrlee, line 57:
```suggestion
    if (error == nil) {
        self.requests += 1;
        return;
    }
    // content of else block...
    
```
Btw, what's the guideline around `self.requests += 1;` vs `self.requests++;`? are both allowed/idiomatic?

MapboxMobileEvents/MMEMetricsManager.m
```diff
@@ -0,0 +1,160 @@
...
+        
+            NSNumber *failedCount = [self.failedRequestsDict objectForKey:failedRequestKey];
+            failedCount = [NSNumber numberWithInteger:[failedCount integerValue] + 1];
+            [self.failedRequestsDict setObject:failedCount forKey:failedRequestKey];
+        }
+    }
+}
+
+- (void)metricsFromData:(NSData *)data {
+    self.totalDataTransfer = self.totalDataTransfer + data.length;
```
andrlee:
```suggestion
    self.totalDataTransfer += data.length;
```
would this be cleaner? same question on line 90, 92, 97

MapboxMobileEvents/MMEEventsManager.m
```diff
@@ -308,6 +315,35 @@ - (void)sendTurnstileEvent {
     }];
 }
 
+- (void)sendTelemetryMetricsEvent {
```
andrlee, line 327: this method is a bit hard to read. is there a way to split it into the multiple functions?

MapboxMobileEvents/MMEAPIClient.m
```diff
@@ -75,6 +77,11 @@ - (void)postEvents:(NSArray *)events completionHandler:(nullable void (^)(NSErro
             if (completionHandler) {
                 error = error ?: statusError;
                 completionHandler(error);
+
```
andrlee, line 80: it looks like we're violating DRY with this block of code. should we consider wrapping it with helper method?

andrlee (changes requested):

MapboxMobileEvents/MMEAPIClient.m
julianrex: Agreed.

MapboxMobileEvents/MMEConstants.h
```diff
@@ -17,6 +17,7 @@ extern NSString * const MMEAPIClientHTTPMethodGet;
 extern NSString * const MMEErrorDomain;
 
 // Debug types
+extern NSString * const MMEDebugEventType;
```
julianrex, line 20: We should make it clear this is a key; perhaps `MMEDebugEventTypeKey` (or whatever is consistent with existing keys).

MapboxMobileEvents/MMEConstants.m
```diff
@@ -73,16 +77,33 @@
...
+NSString * const MMEEventEventCountTotal = @"eventCountTotal";
+NSString * const MMEEventEventCountMax = @"eventCountMax";
+NSString * const MMEEventDeviceLat = @"deviceLat";
+NSString * const MMEEventDeviceLon = @"deviceLon";
+NSString * const MMEEventDeviceTimeDrift = @"deviceTimeDrift";
+NSString * const MMEEventConfigResponse = @"configResponse";
+
 NSString * const MMEVisionEventPrefix = @"vision.";
+
+NSString * const MMENavigationEventPrefix = @"navigation.";
```
julianrex, line 102: Q. What is prefix used for?

MapboxMobileEvents/MMEEvent.m
```diff
@@ -12,6 +12,17 @@ + (instancetype)turnstileEventWithAttributes:(NSDictionary *)attributes {
     return turnstileEvent;
 }
 
++ (instancetype)telemetryMetricsEventWithDateString:(NSString *)dateString attributes:(NSDictionary *)attributes {
+    MMEEvent *telemetryMetrics = [[MMEEvent alloc] init];
```
julianrex, line 16: It feels like we need an `MMEEvent` initializer that takes a name and attributes parameter.

MapboxMobileEvents/MMEEventsManager.m
```diff
@@ -65,6 +67,7 @@ - (instancetype)init {
         _application = [[MMEUIApplicationWrapper alloc] init];
         _dateWrapper = [[MMENSDateWrapper alloc] init];
         _dispatchManager = [[MMEDispatchManager alloc] init];
+        _metricsManager = [MMEMetricsManager sharedManager];
```
julianrex, line 70: What's different about `MMEMetricsManager` that it's a singleton, rather than being allocated like the others here?

MapboxMobileEvents/MMEMetricsManager.h
julianrex: Yep, let's use `CLLocationDegrees`.

MapboxMobileEvents/MMEMetricsManager.h
julianrex: Also what's the reasoning for having `int` and `long`? Any reason we couldn't use `NSInteger`? (or `NSUInteger` to @andrlee's point)

MapboxMobileEvents/MMEMetricsManager.m
julianrex: `self.requests++` is fine here. (However the increment/decrement operators are deprecated in Swift)

MapboxMobileEvents/MMEMetricsManager.m
```diff
@@ -0,0 +1,160 @@
...
+        self.requests = self.requests + 1;
+    } else {
+        if (events) {
+            self.eventCountFailed = self.eventCountFailed + (int)events.count;
+        }
+        
+        if ([error.userInfo objectForKey:MMEResponseKey]) {
+            NSHTTPURLResponse *response = (NSHTTPURLResponse *)[error.userInfo objectForKey:MMEResponseKey];
+            NSString *urlString = [[response URL] absoluteString];
+            NSNumber *statusCode = [NSNumber numberWithLong:(long)response.statusCode];
```
julianrex: Why the `long`? Also, you could try `@(response.statusCode)` - then you fetch it with the `integerValue` property on `NSNumber`

MapboxMobileEvents/MMEMetricsManager.m
```diff
@@ -0,0 +1,160 @@
...
+
+- (void)captureConfigurationJSON:(NSDictionary *)configuration {
+    if (configuration) {
+        self.configResponseDict = configuration;
+    }
+}
+
+- (void)captureLatitude:(float)lat longitude:(float)lon {
+    if (!self.deviceLat || !self.deviceLon) {
+        self.deviceLat = [[NSString stringWithFormat:@"%.06f", lat] floatValue];
```
julianrex: What's going on here? Looks like we're truncating the precision? Is this for privacy reasons?

MapboxMobileEvents/MMEMetricsManager.m
```diff
@@ -0,0 +1,160 @@
...
+    self.dateUTC = [NSDate date];
+    NSDateFormatter *dateFormatter = [[NSDateFormatter alloc] init];
+    NSTimeZone *utcTimeZone = [NSTimeZone timeZoneWithName:@"UTC"];
+    [dateFormatter setTimeZone:utcTimeZone];
+    [dateFormatter setDateFormat:@"yyyy-MM-dd"];
+    self.dateUTCString = [dateFormatter stringFromDate:self.dateUTC];
+}
+
+- (void)resetMetrics {
+    self.failedRequestsDict = [[NSMutableDictionary alloc] init];
```
julianrex: Perhaps some kind of metrics object would be useful here, so you can just allocate a new one?

MapboxMobileEvents/MMEConstants.h
julianrex: Also, while I think of it - for all these constants we should look at `NS_TYPED_ENUM` - for some additional type safety (and bridging to Swift)

MapboxMobileEvents/MMEMetricsManager.m
rclee: Ahh, you're right -- I had thought this was decremented on ObjC as well 🤦‍♂️ (+= would have been better even so). Thanks!

MapboxMobileEvents/MMEMetricsManager.m
rclee: Yes!

MapboxMobileEvents/MMEEventsManager.m
rclee: Will split this into a method for generating the `MMEEvent`

MapboxMobileEvents/MMEMetricsManager.h
julianrex: See my comment below :)

MapboxMobileEvents/MMEConstants.h
rclee: Good idea -- I think this will have to be added a bit later as a minor version change since it may be changing the way these constants are used.

MapboxMobileEvents/MMEConstants.m
rclee: Used to give our clients early warnings that their event type will not be accepted.

MapboxMobileEvents/MMEEvent.m
rclee: As we discussed this would also be a breaking change and would likely need to be applied at a later stage. This could be useful for some of our plans regarding time drift however.

rclee:
All the requested changes should be in now. Would be great to have if y'all could take another look. Thanks for the suggestions! 

cc: @andrlee @julianrex

MapboxMobileEvents/MMEAPIClient.m
julianrex: Looks like this still needs to be addressed?

MapboxMobileEvents/MMEAPIClient.m
```diff
@@ -75,6 +77,11 @@ - (void)postEvents:(NSArray *)events completionHandler:(nullable void (^)(NSErro
             if (completionHandler) {
                 error = error ?: statusError;
                 completionHandler(error);
+              
+                [self.metricsManager metricsFromEvents:events error:error];
```
julianrex: Something I didn't pick up before - it's not particularly clear from the method name, what's going on here?
As it stands is sounds like something is going to be returned from `metricsFrom...:error:` - what does it do?

MapboxMobileEvents/MMEConstants.m
```diff
@@ -25,14 +27,16 @@
 NSString * const MMEDebugEventTypeBackgroundTask = @"backgroundTask";
 NSString * const MMEDebugEventTypeMetricCollection = @"metricCollection";
 NSString * const MMEDebugEventTypeLocationManager = @"locationManager";
+NSString * const MMEDebugEventTypeTelemetryMetrics = @"telemMetrics";
```
julianrex, line 30: What's the difference between `telemMetrics` and `telemetryMetrics`? I'm confused by that.

MapboxMobileEvents/MMEEvent.m
julianrex: Would having an additional `MMEEvent` initializer be a breaking change?

MapboxMobileEvents/MMEMetricsManager.m
```diff
@@ -0,0 +1,143 @@
...
+
+@implementation MMEMetricsManager
+
++ (instancetype)sharedManager {
+    static MMEMetricsManager *_sharedManager;
+    static dispatch_once_t onceToken;
+    
+    dispatch_once(&onceToken, ^{
+        _sharedManager = [[MMEMetricsManager alloc] init];
+        _sharedManager.metrics = [[MMEMetrics alloc] init];
```
julianrex: Looks like you're already allocating a `MMEMetrics` below - is this necessary here?

MapboxMobileEvents/MMEAPIClient.m
julianrex: Perhaps it should be something like `updateMetricsWithEvents:error`?

MapboxMobileEvents/MMEAPIClient.m
```diff
@@ -104,6 +111,11 @@ - (void)postMetadata:(NSArray *)metadata filePaths:(NSArray *)filePaths completi
         if (completionHandler) {
```
andrlee, line 100: while the block of code (line 111-118) deserves to live in its own method. there's one more thing that can improve readability. Consider another block (line 100-109): could this be simplified to:
```
NSError *statusError = [self getStatusError:response];
```
and the helper method might look like
```
(NSError *) getStatusError: (NSHTTPURLResponse *)response {
   if (response.statusCode != 200) {
       return nil;
   } 
   .....
}
```
Please forgive me for mistakes in syntax, i promise i'll become better at objective-c over the time...

MapboxMobileEvents/MMEAPIClient.m
```diff
@@ -104,6 +111,11 @@ - (void)postMetadata:(NSArray *)metadata filePaths:(NSArray *)filePaths completi
...
             error = error ?: statusError;
             completionHandler(error);
+
```
andrlee, line 103: DRY

MapboxMobileEvents/MMEEventsManager.m
```diff
@@ -308,6 +315,42 @@ - (void)sendTurnstileEvent {
     }];
 }
 
+- (MMEEvent *)generateTelemetryMetricsEvent {
```
andrlee: it looks like `MMEEventsManager` will soon become unmanageable if we continue adding `generateFooEvent` and `sendFooEvent` with every new event type. Should we consider refactoring this part? One option would be to create a generic interface that exposes one method: `createEvent` in which you can combine both creation and validation logic. It seems like sending can be simplified to just posting with generic callback that handles result of the network request.

MapboxMobileEvents/MMEMetricsManager.m
```diff
@@ -0,0 +1,143 @@
...
+- (instancetype)init {
+    self = [super init];
+    if (self) {
+        _metrics = [[MMEMetrics alloc] init];
+    }
+    return self;
+}
+
+- (void)metricsFromEventQueue:(NSArray *)eventQueue {
+    if (self.metrics.dateUTC == nil) {
```
andrlee: this block line 35-37 conflicts with the name of this method...

andrlee (changes requested):

MapboxMobileEvents/MMEMetricsManager.m
```diff
@@ -0,0 +1,143 @@
...
+
+- (void)captureCoordinate:(CLLocationCoordinate2D)coordinate; {
+    if (!self.metrics.deviceLat || !self.metrics.deviceLon) {
+        self.metrics.deviceLat = round(coordinate.latitude*1000)/1000;
+        self.metrics.deviceLon = round(coordinate.longitude*1000)/1000;
+    }
+}
+
+- (void)updateDateUTC {
+    self.metrics.dateUTC = [NSDate date];
```
julianrex: The `dateUTC` property should probably just be called `date` - you use the date formatter to create the UTC string (though it could be any time zone for that matter)

MapboxMobileEvents/MMEAPIClient.m
rclee: Sorry, seems like I misinterpreted what was desired here before. Seems like the portion that shouldn't be repeated is:

```
[self.metricsManager metricsFromEvents:filePaths error:error]; 
if (request.HTTPBody && error == nil) {
    [self.metricsManager metricsFromData:request.HTTPBody];
}
```

I'll move this into the above method and add a parameter for the `request` conditional.

MapboxMobileEvents/MMEAPIClient.m
rclee: No worries! This helper method actually exists below
 (`[self statusErrorFromRequest:andHTTPResponse:`) but during some merges it seems that it was used in some methods but not here. Thanks for the sample and pointing that out.

MapboxMobileEvents/MMEMetricsManager.m
rclee: It shouldn't be necessary. I had added the `[super init]` below for testing and forgot I had this up here.

MapboxMobileEvents/MMEConstants.m
rclee: `telemMetrics` is only used for internal debugging purposes. `telemetryMetrics` is used when posting the actual event to the server.

MapboxMobileEvents/MMEAPIClient.m
rclee: Good point -- should have caught that.

MapboxMobileEvents/MMEEventsManager.m
```diff
@@ -202,6 +205,10 @@ - (void)flush {
     
     NSArray *events = [self.eventQueue copy];
     NSUInteger eventsCount = events.count;
+
```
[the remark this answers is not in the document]
rclee: I'd argue that cleaning the queue should still be part of `flush()` otherwise it would only be posting the events. Also, for debug events -- these are already being called on a separate method. I think inline is the cleanest way to do this and provides some additional documentation for developers freshly entering this project. 

I agree this method has gotten a bit bloated. I'll see how we can make this more readable. Thanks for bringing this up 👍

MapboxMobileEvents/MMEMetricsManager.m
rclee: Why would the time zone differ? According the the [NSDate docs](https://developer.apple.com/documentation/foundation/nsdate?language=objc) `[NSDate date]` should always reference UTC timezone until formatted and then it would be changed to the desired timezone.

MapboxMobileEvents/MMEMetricsManager.m
rclee: Good point, moved `date` to be set when `MMEMetrics` is initialized and formatted when `attributes` are requested.

MapboxMobileEvents/MMEAPIClient.m
rclee: Yeah, sorry. I had misinterpreted this to be referring to something else. Should be fixed now

MapboxMobileEvents/MMEEvent.m
rclee: We should be able do that. Have a convenience initializer that takes in the `name`, `dateString`, and additional `attributes`. Later we can look at getting the `dateString` to be an actual `date` that will be easier to manipulate.

```
MMEEvent *telemetryMetrics = [[MMEEvent alloc] initWithName:name created:dateString attributes:attributes];
```

julianrex: Perhaps we hold off on this for when we consider throwing around `NSDate`s? Otherwise we might end up having to remove an initializer later. Up to you :)

rclee: Well, either way. The initializer should take the date as a formatted string to match our schema. Any magic would have to happen before we initialize the event in this case -- which is fine but will probably be a major version change at that point since we'll require an NSDate as a parameter for this method. 

The bad thing about this change is that it would put more pressure on the MME library to keep the date format matching the schema. Which is something we were trying to move away from.

rclee: Would it make sense to pass in an `NSDateFormatter` as a parameter? Then whatever desired format could be used when the event is generated.

julianrex: Idle thought - should the date format be supplied by the configuration api?

rclee: Possibly... but if there ever came a time where an event required a separate date format for whatever reason then we may need to setup an additional configuration for specific events. 

IMO adding a `NSDateFormatter` parameter to our `MMEEvent` constructors makes the most sense for our current setup. This would only require a small change for our clients and would provide us with more flexibility to adjust any dates as necessary. Still, I think this would be considered a major release and we should hold off until we're ready for that. We could look at building out a more complex system to provide common type definitions through a configuration at a later time if we can find more benefit in doing so.

julianrex: 👍

andrlee (approved):

MapboxMobileEvents/MMEMetrics.h
```diff
@@ -0,0 +1,27 @@
...
+@property (nonatomic) NSUInteger requests;
+@property (nonatomic) NSUInteger totalDataTransfer;
+@property (nonatomic) NSUInteger cellDataTransfer;
+@property (nonatomic) NSUInteger wifiDataTransfer;
+@property (nonatomic) NSUInteger appWakeups;
+@property (nonatomic) NSUInteger eventCountFailed;
+@property (nonatomic) NSUInteger eventCountTotal;
+@property (nonatomic) NSUInteger eventCountMax;
+@property (nonatomic) NSInteger deviceTimeDrift;
+@property (nonatomic) CLLocationDegrees deviceLat;
```
julianrex, line 17: Ideally we should replace `deviceLat` and `deviceLon` by a single `deviceCoordinate`.

MapboxMobileEvents/MMEMetricsManager.m
```diff
@@ -0,0 +1,170 @@
...
+    self.metrics.appWakeups++;
+}
+
+- (void)updateConfigurationJSON:(NSDictionary *)configuration {
+    if (configuration) {
+        self.metrics.configResponseDict = configuration;
+    }
+}
+
+- (void)updateCoordinate:(CLLocationCoordinate2D)coordinate; {
```
julianrex, line 102: We could move this logic into `MMEMetrics`

julianrex (approved):
Couple of minor things - we could address these later, since it's an internal type.

PR #90, opened 2018-11-12
Prepare to publish metrics to AWS

rclee:
Adds and edits files in preparation to send metrics to AWS through CI.


scripts/check_binary_size.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+
+set -e
+set -o pipefail
+
+file_path="build/mapbox-events-ios-static.zip"
+file_size=$(wc -c <"$file_path" | sed -e 's/^[[:space:]]*//')
+date=`date '+%Y-%m-%d'`
+utc_iso_date=`date -u +'%Y-%m-%dT%H:%M:%SZ'`
+label="Telemetry Static ZIP"
```
andrlee: @rclee i'm not sure if we need to measure archive size. can we start with just static lib?

andrlee (approved):

PR #93, opened 2019-01-03
Provide whitelisted event initializer

rclee:
This PR will add a generic event initializer to allow for faster iterations of new events. The main purpose of bringing this out now is our performance push and the need for many new events to be added and sent quickly.

Caution should be used when using this initializer. Such as enabling debug mode on `eventsManager` and ensuring your events make it to the server. Failure to do so could result in a loss of telemetry data.

### DIFF
--- a/MapboxMobileEvents/MMEEvent.m
+++ b/MapboxMobileEvents/MMEEvent.m
@@ -160,6 +160,17 @@
     return carplayEvent;
 }
 
++ (instancetype)eventWithDateString:(NSString *)dateString name:(NSString *)name attributes:(NSDictionary *)attributes {
+    MMEEvent *event = [[MMEEvent alloc] init];
+    event.name = name;
+    NSMutableDictionary *commonAttributes = [NSMutableDictionary dictionary];
+    commonAttributes[MMEEventKeyEvent] = event.name;
+    commonAttributes[MMEEventKeyCreated] = dateString;
+    [commonAttributes addEntriesFromDictionary:attributes];
+    event.attributes = commonAttributes;
+    return event;
+}
+
 + (NSInteger)contentSizeScale {
     NSInteger result = -9999;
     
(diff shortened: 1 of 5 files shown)


MapboxMobileEvents/MMEEvent.m
```diff
@@ -160,6 +160,17 @@ + (instancetype)carplayEventWithName:(NSString *)name attributes:(NSDictionary *
     return carplayEvent;
 }
 
++ (instancetype)eventWithDateString:(NSString *)dateString name:(NSString *)name attributes:(NSDictionary *)attributes {
```
julianrex, line 163: Just a couple of questions here (which I think we've talked about before briefly):

1. Why is this a class method rather than an `MMEEvent` initializer? Is this for future support for different `MMEEvent` subclasses? So does this mean `-[MMEEvent init]` is essentially "private"? If so, I wonder if it's worth having a specific initializer (in addition) and marking the default `init` as `NS_UNAVAILABLE`. Thoughts? 
2. Should this method take an `NSDate` rather than a date string? Perhaps a date transformer could be supplied (somewhere).

rclee: This PR is somewhat of a stopgap measure until we can completely remove the whitelisting that is currently in place. The [doc](https://paper.dropbox.com/doc/iOS-Removing-Whitelist-from-event-queuing--AU~Rjc1CqFqIRDaJTQ7GKL4wAg-Uqob9uEUWBF0t67RUuGTD) describes some of those plans and how we might accomplish them. 

1. I'm not sure how the decision came about to make these class methods or what the future plans were for it. I don't see any harm in a user of the library using an initializer vs any of these class methods to construct an event however. Is there something I'm missing here?
2. I agree that this could be provided in another way and potentially better way (especially once we implement date/time correction {insert Doctor Who meme}) but I think that should be done a little further down the road. For now the date string is supplied by the `eventsManager` via [`MMENSDateWrapper`](https://github.com/mapbox/mapbox-events-ios/blob/master/MapboxMobileEvents/MMENSDateWrapper.m#L27-L29) and all that needs to be called by the client/host is [`enqueueEventWithName:attributes:`](https://github.com/mapbox/mapbox-events-ios/blob/master/MapboxMobileEvents/MMEEventsManager.m#L353-L394).

julianrex: > Is there something I'm missing here?

No - more curious about future plans 😄 

> [...] but I think that should be done a little further down the road

Thanks for clarifying 👍

julianrex (approved):

PR #97, opened 2019-02-08
Add location permissions attributes to turnstile event

rclee:
Adds two new attributes to `turnstile` in an effort to learn more about how location permissions are being used.


MapboxMobileEvents/MMEEventsManager.m
```diff
@@ -591,5 +594,29 @@ - (void)locationManager:(MMELocationManager *)locationManager didVisit:(CLVisit
     }
 }
 
+- (NSString *)locationAuthorizationStatus {
+    CLAuthorizationStatus status = [CLLocationManager authorizationStatus];
+    switch (status) {
+        case kCLAuthorizationStatusDenied:
+            return MMEEventStatusDenied;
+            break;
```
alfwatt: return then break seems awkward, I'd recommend using a local variable to hold the result string, and having a single return (this was SoP at Apple, not sure what the local standard is).

MapboxMobileEvents/MMEEventsManager.m
```diff
@@ -591,5 +594,29 @@ - (void)locationManager:(MMELocationManager *)locationManager didVisit:(CLVisit
     }
 }
 
+- (NSString *)locationAuthorizationStatus {
```
alfwatt: this is a static function, it should be declared static, potentially in a category of CLLocationManager

alfwatt (commented):
Some notes

MapboxMobileEvents/MMEEventsManager.m
rclee: Good catch, been a while since I've used switch!

MapboxMobileEvents/MMEEventsManager.m
rclee: Added in these changes, thanks! Lemme know how it looks now. 🙇

MapboxMobileEvents/CLLocationManager+MMEMobileEvents.m
```diff
@@ -0,0 +1,35 @@
+#import "CLLocationManager+MMEMobileEvents.h"
+#import "MMEConstants.h"
+
+@implementation CLLocationManager (MMEMobileEvents)
+
+void mme_linkCLLocationManagerCategory(){}
+
++ (NSString *)locationAuthorizationStatus {
```
alfwatt: suggest renaming to `mme_authorizationStatusString` to prevent potential namespace collision and match to the underlying call name more closely

alfwatt (dismissed):
Suggest change to category method name

MapboxMobileEvents/CLLocationManager+MMEMobileEvents.m
rclee: Good point

alfwatt (approved):

PR #98, opened 2019-02-08
Enable Link Time Optimization (LTO) compiler flag

friedbunny:
Enables [Link Time Optimization](https://llvm.org/docs/LinkTimeOptimization.html) (LTO) for all targets, resulting in marginally smaller binary sizes and (theoretical) performance improvements.

- LTO increases the amount of time during the build process spent linking, but for a project this size it’s negligible — on my computer, it increased from 1.1 seconds to 1.9 seconds.
- [ThinLTO](https://clang.llvm.org/docs/ThinLTO.html) did not show a measurable decrease in binary size, unfortunately, but “monolithic” did (and is what this PR enables).
- This change should provide improvements for consumer projects/libraries that build this Xcode project directly or consume it as a binary.
- There will be no change in the iOS Maps SDK — https://github.com/mapbox/mapbox-gl-native/pull/12502 already did this (and, due how to this events library is consumed there, it is also LTO'd).

## Binary size changes

The differences in the stripped binaries for the dynamic framework are not huge, but they’re something.

### Before, LTO disabled
```
3.2M	MapboxMobileEvents
496K	MapboxMobileEvents-stripped
272K	MapboxMobileEvents-stripped-arm64
240K	MapboxMobileEvents-stripped-armv7
```

### Monolithic LTO enabled
```
2.8M	MapboxMobileEvents
480K	MapboxMobileEvents-stripped
272K	MapboxMobileEvents-stripped-arm64
224K	MapboxMobileEvents-stripped-armv7
```

### ThinLTO enabled
```
3.2M	MapboxMobileEvents
496K	MapboxMobileEvents-stripped
272K	MapboxMobileEvents-stripped-arm64
240K	MapboxMobileEvents-stripped-armv7
```

/cc @rclee @andrlee


friedbunny:
Looks like this change decreases the static library’s raw binary size from [8.54 MB](https://github.com/mapbox/mapbox-events-ios/runs/50889954) to 7.06 MB.

rclee (approved):
LGTM, Thanks Jason!
cc: @alfwatt @andrlee

PR #100, opened 2019-02-13
JSON serialization error handling

alfwatt:
Needs tests

### DIFF
--- a/MapboxMobileEvents/MMEAPIClient.m
+++ b/MapboxMobileEvents/MMEAPIClient.m
@@ -16,7 +16,7 @@ typedef NS_ENUM(NSInteger, MMEErrorCode) {
 #pragma mark -
 
 @interface MMEEventsManager (Private)
-- (void) pushDebugEventWithAttributes:(NSDictionary*) debugAttributes;
+- (void)pushEvent:(MMEEvent *)event;
 @end
 
 #pragma mark -
@@ -244,12 +244,7 @@ int const kMMEMaxRequestCount = 1000;
             [request setHTTPBody:jsonData];
         }
     } else if (jsonError) {
-        [[MMEEventsManager sharedManager] pushDebugEventWithAttributes:@{
-            MMEDebugEventType: MMEDebugEventTypeError,
-            MMEEventKeyErrorCode: @(jsonError.code),
-            MMEEventKeyErrorDescription: jsonError.localizedDescription,
-            MMEEventKeyErrorFailureReason: jsonError.localizedFailureReason
-        }];
+        [[MMEEventsManager sharedManager] pushEvent:[MMEEvent debugEventWithError:jsonError]];
 
         return nil;
     }
@@ -284,12 +279,7 @@ int const kMMEMaxRequestCount = 1000;
         [httpBody appendData:jsonData];
         [httpBody appendData:[[NSString stringWithFormat:@"\r\n\r\n"] dataUsingEncoding:NSUTF8StringEncoding]];
     } else if (jsonError) {
-        [[MMEEventsManager sharedManager] pushDebugEventWithAttributes:@{
-            MMEDebugEventType: MMEDebugEventTypeError,
-            MMEEventKeyErrorCode: @(jsonError.code),
-            MMEEventKeyErrorDescription: jsonError.localizedDescription,
-            MMEEventKeyErrorFailureReason: jsonError.localizedFailureReason
-        }];
+        [[MMEEventsManager sharedManager] pushEvent:[MMEEvent debugEventWithError:jsonError]];
     }
 
     for (NSString *path in filePaths) { // add a file part for each
--- a/MapboxMobileEvents/MMEAPIClient.m
+++ b/MapboxMobileEvents/MMEAPIClient.m
@@ -4,6 +4,7 @@
 #import "MMEEvent.h"
 #import "NSData+MMEGZIP.h"
 #import "MMEMetricsManager.h"
+#import "MMEEventsManager.h"
 
 typedef NS_ENUM(NSInteger, MMEErrorCode) {
     MMESessionFailedError,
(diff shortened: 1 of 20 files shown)


rclee (dismissed):
Thanks and congrats on PR #100 🎉

rclee:
Maybe we should add tests before we merge though

alfwatt:
Added some simple tests to make sure Error and Exception events are only logged locally, not queued for upload

MapboxMobileEventsTests/MMEEventsManagerTests.mm
```diff
@@ -962,7 +964,36 @@ @interface MMEEventsManager (Tests) <MMELocationManagerDelegate>
...
+            eventsManager.paused = YES;
+        });
+
+        context(@"when an error event is pushed", ^{
+            NSError* testError = [NSError.alloc initWithDomain:NSCocoaErrorDomain code:999 userInfo:@{
+                NSLocalizedDescriptionKey: @"Test Error Description",
+                NSLocalizedFailureReasonErrorKey: @"Test Error Failure Reason"
+            }];
+
+            [[MMEEventsManager sharedManager] pushEvent:[MMEEvent debugEventWithError:testError]];
```
akitchen: You’ll want to put this inside a `beforeEach`, `it`, or `subjectAction` block for a more expected execution order. Otherwise it will be run when the spec is loaded, which might be too early. Same thing below.

akitchen (commented):
It’s fine to declare value objects for setup or expected results in a `context` or `describe` block (which are the same thing BTW) but the action portion of your tests should probably execute later, via a `beforeEach`, `it`, or `subjectAction` block.

akitchen:
LGTM, especially from the testing perspective. I'll leave final approval up to the team 👍

MapboxMobileEvents/MMEAPIClient.m
```diff
@@ -217,16 +230,23 @@ - (NSURLRequest *)requestForEvents:(NSArray *)events {
...
-        NSData *compressedData = [jsonData mme_gzippedData];
-        [request setValue:@"gzip" forHTTPHeaderField:MMEAPIClientHeaderFieldContentEncodingKey];
-        [request setHTTPBody:compressedData];
-    } else {
-        [request setValue:nil forHTTPHeaderField:MMEAPIClientHeaderFieldContentEncodingKey];
-        [request setHTTPBody:jsonData];
+    NSError* jsonError = nil;
+    NSData *jsonData = [NSJSONSerialization dataWithJSONObject:eventAttributes options:0 error:&jsonError];
+
+    if (jsonData) { // we were able to convert the eventAttrubutes, attempt compresison
```
rclee, line 236: nit: "eventAttributes" and "compression"

rclee (approved):
Small typo -- everything else looks solid. Thanks Alf! 👍

MapboxMobileEvents/MMEAPIClient.m
```diff
@@ -66,21 +77,23 @@ - (void)postEvents:(NSArray *)events completionHandler:(nullable void (^)(NSErro
...
+                NSError *statusError = nil;
+                if ([response isKindOfClass:[NSHTTPURLResponse class]]) {
+                    NSHTTPURLResponse *httpResponse = (NSHTTPURLResponse *)response;
+                    statusError = [self statusErrorFromRequest:request andHTTPResponse:httpResponse];
+                } else {
+                    statusError = [self unexpectedResponseErrorfromRequest:request andResponse:response];
+                }
+                if (completionHandler) {
+                    error = error ?: statusError;
+                    completionHandler(error);
```
julianrex, line 91: `completionHandler` will get called for each batch - is that correct?

Seems like it should only be called once? If so we might want to consider a `dispatch_group` here.

MapboxMobileEvents/MMEAPIClient.m
```diff
@@ -66,21 +77,23 @@ - (void)postEvents:(NSArray *)events completionHandler:(nullable void (^)(NSErro
...
+
+                    [self.metricsManager updateMetricsFromEvents:events request:request error:error];
```
julianrex, line 93: Looks like this should be outside the `if (completionHandler)` block?
Also should `updateMetricsFromEvents:request:` be passed `batch` instead of `events`?

MapboxMobileEvents/MMEAPIClient.m
```diff
@@ -66,21 +77,23 @@ - (void)postEvents:(NSArray *)events completionHandler:(nullable void (^)(NSErro
...
+                }
+            }];
+        }
```
julianrex, line 96: If request is nil (and the completion block is called for each batch) should it be called here too?

julianrex (commented):
Not really a review, just something I spotted that looked a bit off (also in the original code). I may be missing context, so apologies if I'm mistaken here!

MapboxMobileEvents/MMEAPIClient.m
rclee: Thanks for looking into this @julianrex! I'll try to give some more context behind these decisions. 

We wanted to prepare for deferred telemetry by implementing batching. Each batch represents the estimated event limit (around 1000) when sending a request. This should allow for us to send as many events as we desire from local storage and let `MMEAPIClient` handle the sending of those events. So, it seemed to me that each batch should be a separate call and return separately. Let me know if that makes sense.

MapboxMobileEvents/MMEAPIClient.m
rclee: Good catch. That should be the case. Passing in the batch shouldn't matter for our current setup but it certainly will for deferred telemetry.

MapboxMobileEvents/MMEAPIClient.m
rclee: Yes, the events/batches should be counted in a non-deferred state in this case. 👍 

I'll start a new PR for these changes. Thanks!

MapboxMobileEvents/MMEAPIClient.m
julianrex: Thanks for providing context, and apologies for leaping in without fully knowing the background 😉

Sounds like you need a callback per batch - a progress update I suppose?

Calling a completion block multiple times is non-standard and confusing (given that it's call "completion"), so perhaps we need two things here:
- a progress block (i.e. what you have so far, though passing in the block or some kind of block id might be useful)
- a completion block for when all the blocks have been posted.

MapboxMobileEvents/MMEAPIClient.m
julianrex: np!

PR #74, opened 2018-10-02
TrustKit fix

rclee:
Changes to TrustKit to fix [nullability warnings](https://github.com/mapbox/mapbox-events-ios/issues/66) in Xcode 10/iOS 12 and fix [a data race](https://github.com/mapbox/mapbox-events-ios/issues/71) with TrustKit's logging. The data race issue is only an issue in non-release builds, in release builds the logger is disabled.

PR #101, opened 2019-02-19
Fix tests

rclee:
We need to replace `didRecieveChallenge` related test since we moved directions. Comparing the received disposition from the handler is more valuable. 

ref: https://github.com/mapbox/mapbox-events-ios/pull/74

### DIFF
--- a/MapboxMobileEventsTests/MMEAPIClientTests.mm
+++ b/MapboxMobileEventsTests/MMEAPIClientTests.mm
@@ -30,10 +30,11 @@ SPEC_BEGIN(MMEAPIClientSpec)
 describe(@"MMEAPIClient", ^{
     
     __block MMEAPIClient *apiClient;
-    __block int64_t timeoutInNanoseconds = 1000000000;
     __block NSURLSessionAuthChallengeDisposition receivedDisposition;
-    __block NSURLSessionAuthChallengeDisposition expectedDefaultDisposition = NSURLSessionAuthChallengePerformDefaultHandling;
-    __block NSURLSessionAuthChallengeDisposition expectedCancelDisposition = NSURLSessionAuthChallengeCancelAuthenticationChallenge;
+    
+    int64_t timeoutInNanoseconds = 1000000000;
+    NSURLSessionAuthChallengeDisposition expectedDefaultDisposition = NSURLSessionAuthChallengePerformDefaultHandling;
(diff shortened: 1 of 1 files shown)


MapboxMobileEventsTests/MMEAPIClientTests.mm
```diff
@@ -31,6 +31,9 @@ - (void)setupUserAgent;
     
     __block MMEAPIClient *apiClient;
     __block int64_t timeoutInNanoseconds = 1000000000;
+    __block NSURLSessionAuthChallengeDisposition receivedDisposition;
+    __block NSURLSessionAuthChallengeDisposition expectedDefaultDisposition = NSURLSessionAuthChallengePerformDefaultHandling;
```
akitchen: `__block` is only needed for references which will be assigned to (changed) within a block other than the current scope. `timeoutInNanoseconds`, `expectedDefaultDisposition`, and `expectedCancelDisposition` technically don't require `__block`

MapboxMobileEventsTests/MMEAPIClientTests.mm
```diff
@@ -84,19 +87,15 @@ - (void)setupUserAgent;
...
             });
             
-            it(@"should receive challenge", ^{
-                apiClient.sessionWrapper should have_received(@selector(URLSession:didReceiveChallenge:completionHandler:));
-            });
-            
-            it(@"should not receive challenge", ^{
-                apiClient.sessionWrapper should_not have_received(@selector(URLSession:task:didReceiveChallenge:completionHandler:));
+            it(@"should equal expected CancelAuthentication disposition", ^{
+                expectedCancelDisposition should equal(receivedDisposition);
```
akitchen: I would write this assertion the other way around

akitchen: same thing below...

akitchen:
I think this is a good improvement, but ultimately we care about testing the results of 

https://github.com/mapbox/mapbox-events-ios/blob/5075b42512ca8675a77b8fd85f0f629eb22c93c3/MapboxMobileEvents/MMENSURLSessionWrapper.m#L73

which, to do right (and to eliminate the possible need for hitting the network during test runs) we should be injecting a fake `trustKit.pinningValidator` which we can use to simulate/control the results of `-[pinningValidator handleChallenge:completionHandler:]` in the tests

akitchen (dismissed):

alfwatt (approved):
Implemented review comments

Issue #91, opened 2018-11-20
Fix variety of minor warnings


friedbunny:
Enabling `WARNING_CFLAGS = -Wunused-property-ivar -Wsemicolon-before-method-body -Wmismatched-tags -Wsign-compare` gets us a variety of warnings we should fix:

```
mapbox-events-ios/MapboxMobileEvents/MMEMetricsManager.m:102:60: Semicolon before method body is ignored
mapbox-events-ios/vendor/TrustKit/Reporting/TSKBackgroundReporter.m:41:77: Semicolon before method body is ignored
mapbox-events-ios/vendor/TrustKit/Reporting/TSKPinFailureReport.m:57:25: Semicolon before method body is ignored
mapbox-events-ios/MapboxMobileEvents/MMEEvent.m:54:120: Semicolon before method body is ignored
mapbox-events-ios/MapboxMobileEvents/MMEDependencyManager.m:16:47: Semicolon before method body is ignored

mapbox-events-ios/MapboxMobileEvents/MMEEventLogger.m:55:1: Ivar '_timeForNewLogFile' which backs the property is not referenced in this property's accessor
```

These are all trivial issues, but several of them exist in TrustKit, so we’ll have to decide on a strategy for how to fix/handle those.

/cc @rclee @andrlee @zugaldia @chloekraw

alfwatt:
Same errors as: https://github.com/mapbox/mobile-telemetry/issues/283

PR #102, opened 2019-02-19
Fix minor warnings

rclee:
Fixes https://github.com/mapbox/mapbox-events-ios/issues/91 but we still need a better solution for keeping TrustKit up-to-date.


alfwatt (approved):
Low risk changes, approved

PR #108, opened 2019-03-04
Correct SessionWrapper unit tests

rclee:
Some of the tests didn't use proper tooling and were all in all a bit messy. 

This commit should clean up the unit tests around the session wrapper and stubs the `handlechallenge` method to ensure that it returns the correct value for unit testing. The test we had before (testing `NSURLSessionAuthChallengeCancelAuthenticationChallenge` may be added back as an integration test). 

LLVM_LTO is also disabled for debug modes. This is to allow breakpoints to operate properly. `-fcxx-modules` and `-fmodules` are added to import #TrustKit into the .mm file.

Co-authored-by: Andrew-Kitchen <andrew.kitchen@mapbox.com>


akitchen:
The `context` at line 78 and line 147 seem to be the same thing, since the tests themselves run synchronously on the main queue. Not sure if you need to keep both of these around. Other than that, LGTM :)

akitchen (dismissed):

MapboxMobileEvents.xcodeproj/project.pbxproj
```diff
@@ -1171,6 +1171,7 @@
 				INSTALL_PATH = "$(LOCAL_LIBRARY_DIR)/Frameworks";
 				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks @loader_path/Frameworks";
+				LLVM_LTO = NO;
```
akitchen, line 1167: Were these changes in LTO intentional? Not sure if they're actually needed...

rclee: Yeah, I needed to turn off LTO on debug builds -- they were preventing breakpoints from operating for some reason.

rclee:
Implemented your suggestions and removed the setting of the TrustKit Logger which is only used in debug mode. The Logger was removed to fix an error on nightly builds (we're not using the Logger anyway).

alfwatt (dismissed):
Changes look good, make sure to rebase the branch, the `LLVM_LTO` change might conflict (or be overrides by) changes on `master` which disabled it for all targets in the project

Issue #110, opened 2019-03-05
Report size of stripped static library when building


alfwatt:
In order to track our binary release size over time we are reporting it as part of the build process, currently the build script reports the size of the debug build of the static library.

Update the build scripts to report the stripped, optimized size of the static library.

alfwatt:
Defer to 1.0.0

PR #112, opened 2019-03-06
Issue #110 — Report size of stripped static library only when merging to master

alfwatt:
Changes are in the `bitrise.yml` config, this PR should not have size information added until it's merged.


alfwatt:
and yet, there it is…

rclee (approved):
Is this why we're missing our binary sizes? 😬

alfwatt:
This is just adding the scripts and docs, we need to update the bitrise.yml I think

PR #113, opened 2019-03-13
telemetryMetrics improvements

rclee:
This PR implements [schema changes](https://github.com/mapbox/event-schema/commit/74f60cdefcdb967ff16f8e61f71f06f90aec5584) for `telemetryMetrics` to begin tracking received data and fixes some bugs with event counting. 

Fixes https://github.com/mapbox/mobile-telemetry/issues/303


MapboxMobileEvents/MMEMetricsManager.h
```diff
@@ -10,9 +10,10 @@ NS_ASSUME_NONNULL_BEGIN
 
 + (instancetype)sharedManager;
 
-- (void)updateMetricsFromData:(NSData *)data;
+- (void)updateTransferredData:(NSData *)data;
+- (void)updateReceivedData:(NSData *)data;
```
alfwatt: As above suggest `NSUInteger bytes` and `updateReceivedBytes`

MapboxMobileEvents/MMEMetricsManager.m
```diff
@@ -49,37 +49,63 @@ - (void)updateMetricsFromEventQueue:(NSArray *)eventQueue {
...
+        
+        NSNumber *failedCount = [failedRequests objectForKey:statusCodeKey];
+        failedCount = [NSNumber numberWithInteger:[failedCount integerValue] + 1];
+        [failedRequests setObject:failedCount forKey:statusCodeKey];
+        
+        [self.metrics.failedRequestsDict setObject:failedRequests forKey:MMEEventKeyFailedRequests];
+    }
+}
+
+- (void)updateEventsFailedCountFromEvents:(NSArray *)events {
```
alfwatt: Suggest `NSUInteger eventCount` and rename to `updateEventsFailedCount`

MapboxMobileEvents/MMEMetrics.h
```diff
@@ -9,6 +9,9 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic) NSUInteger totalDataTransfer;
```
alfwatt: Suggest making `readonly`, renaming `totalBytesTransferred` and computing from `cellBytesTransfered + wifiBytesTransfered`

MapboxMobileEvents/MMEMetrics.h
```diff
@@ -9,6 +9,9 @@ NS_ASSUME_NONNULL_BEGIN
...
 @property (nonatomic) NSUInteger cellDataTransfer;
```
alfwatt: Suggest making `readonly`, renaming to `cellBytesTransferred` and computing from `cellBytesSent + cellBytesRecieved`

MapboxMobileEvents/MMEMetrics.h
```diff
@@ -9,6 +9,9 @@ NS_ASSUME_NONNULL_BEGIN
...
 @property (nonatomic) NSUInteger wifiDataTransfer;
```
alfwatt: Suggest making `readonly`, renaming to `wifiBytesTransfered` and computing the value

alfwatt (commented):
Suggestions inline, generally: try to be DRY about your data storage, computing the totals makes it easier to keep the test passing and saves a little memory at the expense of a little runtime.

Better to pass in the values, instead of the containers (NSData/NSArray) to prevent accidental retention. Method/property names should reflect the units they're recording (Bytes vs. Data for e.g.)

MapboxMobileEvents/MMEMetrics.m
```diff
@@ -5,24 +5,29 @@ @implementation MMEMetrics
...
+        _eventCountMax = 0;
+        _appWakeups = 0;
+        _deviceLat = 0;
+        _deviceLon = 0;
+        _requests = 0;
     }
     return self;
 }
 
+- (void)computeTransferredBytes {
```
alfwatt: Suggest removing the `_total…` properties and adding:

`- (NSUInteger) totalBytesSent { return _wifiBytesSent + _cellBytesSent; }`

This will satisfy the requirements for the `@property` directive and you won't have to remember to call this compute method, or store the product until it's needed for an event report.

alfwatt (commented):
Sorry I was a little unclear about the `readonly` properties, `computed` might be a better way to describe them than `synthetic` which means something else in Java for e.g.: when the property getter is called, you can do the math real quick, instead of looking at the stored value.

Slightly more compute time but less memory use, and no chance of the values getting out of synch.

MapboxMobileEvents/MMEEventsManager.m
```diff
@@ -187,13 +186,14 @@ - (void)pauseOrResumeMetricsCollectionIfRequired {
...
     } else if (!self.paused && ![self isEnabled]) {
         [self flush];
-        [self sendTelemetryMetricsEvent];
         [self resetEventQueuing];
         [self pauseMetricsCollection];
     }
 }
 
 - (void)flush {
+    [self sendTelemetryMetricsEvent];
```
alfwatt, line 193: Good Catch!

alfwatt (dismissed):
Looks good, just needs the test cases around the timer and we're good to commit!

alfwatt (commented):
Also, you should rebase this branch from master:

`git fetch`
`git rebase master`

There are a couple of small conflicts where I'd renamed `date` to `recordingStarted` that you'll need to resole by hand. if you run into trouble (it can be easy to end up on a detached branch if you don't follow the rebase procedure just so) let me know and we can pair up

alfwatt (approved):

PR #115, opened 2019-03-21
Fix telemetryMetrics bugs and schema attribute name change

rclee:
Fixes a bug with empty dictionaries being added to the event and removes `userId` (which didn't belong there).

Changes dataTransfer to dataSent to reflect changes in events schema


MapboxMobileEvents/MMEMetricsManager.m
```diff
@@ -212,7 +211,8 @@ - (void)pushDebugEventWithAttributes:(NSDictionary *)attributes {
 #pragma mark -
 
 - (NSString *)jsonStringfromDict:(NSDictionary *)dictionary {
-    if (dictionary) {
+    //prevents empty dictionaries from being stringified
+    if ([dictionary count] > 0) {
```
alfwatt, line 215: Do you want `dictionary && dictionary.count > 0` here?

alfwatt (approved):
Never mind the extra null check

PR #120, opened 2019-04-01
Disable TrustKit logging

rclee:
TrustKit logging is causing too much debug logging in the console. Disabling this for now, since we'll soon be removing most of the TrustKit.


alfwatt (approved):
LGTM - rebase and go!

PR #130, opened 2019-05-02
Fix account leak

rclee:
This should fix a memory leak we have associated with MGLAccountManager. Two things needed to be fixed 

1) calling `initializeWithAccessToken: userAgentBase: hostSDKVersion:` too many times could cause memory leaks and potentially other issues.  
2) `MMENSURLSessionWrapper` wasn't being invalidated properly and was causing some memory leaks

ref: https://github.com/mapbox/gl-internal/issues/1126

### DIFF
--- a/MapboxMobileEvents/MMENSURLSessionWrapper.m
+++ b/MapboxMobileEvents/MMENSURLSessionWrapper.m
@@ -1,11 +1,20 @@
 #import "MMENSURLSessionWrapper.h"
+#import "MMEEventsManager.h"
 #import "MMECertPin.h"
 
+#pragma mark -
+
+@interface MMEEventsManager (Private)
+- (void)pushEvent:(MMEEvent *)event;
+@end
+
+#pragma mark -
 
 @interface MMENSURLSessionWrapper ()
 
 @property (nonatomic) dispatch_queue_t serialQueue;
 @property (nonatomic) MMECertPin *certPin;
+@property (nonatomic) NSURLSession *session;
 
 @end
 
@@ -14,6 +23,7 @@
 - (instancetype)init {
     self = [super init];
     if (self) {
+        _session = [NSURLSession sessionWithConfiguration:[NSURLSessionConfiguration defaultSessionConfiguration] delegate:self delegateQueue:nil];
         _serialQueue = dispatch_queue_create([[NSString stringWithFormat:@"%@.events.serial", NSStringFromClass([self class])] UTF8String], DISPATCH_QUEUE_SERIAL);
         _certPin = [[MMECertPin alloc]init];
     }
@@ -24,13 +34,15 @@
     [self.certPin updateWithConfiguration:configuration];
 }
 
+-(void)dealloc {
+    [self.session invalidateAndCancel];
+}
+
 #pragma mark MMENSURLSessionWrapper
 
 - (void)processRequest:(NSURLRequest *)request completionHandler:(void (^)(NSData * _Nullable data, NSURLResponse * _Nullable response, NSError * _Nullable error))completionHandler {
-    NSURLSession *session = [NSURLSession sessionWithConfiguration:[NSURLSessionConfiguration defaultSessionConfiguration] delegate:nil delegateQueue:nil];
-    
     dispatch_async(self.serialQueue, ^{
-        __block NSURLSessionDataTask *dataTask = [session dataTaskWithRequest:request completionHandler:^(NSData * _Nullable data, NSURLResponse * _Nullable response, NSError * _Nullable error) {
+        __block NSURLSessionDataTask *dataTask = [self.session dataTaskWithRequest:request completionHandler:^(NSData * _Nullable data, NSURLResponse * _Nullable response, NSError * _Nullable error) {
             if (completionHandler) {
                 dispatch_async(dispatch_get_main_queue(), ^{
                     completionHandler(data, response, error);
@@ -39,7 +51,6 @@
             dataTask = nil;
         }];
         [dataTask resume];
-        [session finishTasksAndInvalidate];
     });
 }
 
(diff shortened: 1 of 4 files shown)


MapboxMobileEvents/MMENSURLSessionWrapper.m
```diff
@@ -39,6 +39,7 @@ - (void)processRequest:(NSURLRequest *)request completionHandler:(void (^)(NSDat
             dataTask = nil;
         }];
         [dataTask resume];
+        [session finishTasksAndInvalidate];
```
julianrex: If you call `finishTasksAndInvalidate` here won't that finish the dataTask above that was just `resume`d?

MapboxMobileEvents/MMENSURLSessionWrapper.m
```diff
@@ -29,8 +27,10 @@ - (void)reconfigure:(MMEEventsConfiguration *)configuration {
 #pragma mark MMENSURLSessionWrapper
 
 - (void)processRequest:(NSURLRequest *)request completionHandler:(void (^)(NSData * _Nullable data, NSURLResponse * _Nullable response, NSError * _Nullable error))completionHandler {
+    NSURLSession *session = [NSURLSession sessionWithConfiguration:[NSURLSessionConfiguration defaultSessionConfiguration] delegate:nil delegateQueue:nil];
```
julianrex: Do you not need the delegate any more?

MapboxMobileEvents/MMENSURLSessionWrapper.m
rclee: As I understand it, this call will wait until the task finishes and _then_ will invalidate the session object. 

From the docs:
> existing tasks continue until completion. After the last task finishes and the session makes the last delegate call related to those tasks, the session calls the urlSession(_:didBecomeInvalidWithError:) method on its delegate, then breaks references to the delegate and callback objects. After invalidation, session objects cannot be reused.

https://developer.apple.com/documentation/foundation/urlsession/1407428-finishtasksandinvalidate

MapboxMobileEvents/MMENSURLSessionWrapper.m
rclee: I've tested this without the delegate and we're still getting our `URLSession:didReceiveChallenge:` delegate callback. Since the documentation states:

> Your URLSession object doesn’t need to have a delegate. If no delegate is assigned, a system-provided delegate is used, and you must provide a completion callback to obtain the data.

https://developer.apple.com/documentation/foundation/urlsessiondelegate

I didn't see a good reason to keep it around since this is the only delegate method we have at the moment.

rclee:
Moving from a shared session to a session per request is causing some weirdness with the tests that I didn't see last night...

MapboxMobileEvents/MMENSURLSessionWrapper.m
alfwatt: Note that it but doesn't block the caller while waiting: `This method returns immediately without waiting for tasks to finish`

MapboxMobileEvents/MMENSURLSessionWrapper.m
alfwatt: I'm curious why the `system-provided` delegate would continue to call through to our delegate method? How would it event know?

alfwatt (commented):
I'm curious about that delegate case, can we double check with a unit test just to make sure that cert pinning isn't accidentally disabled?

MapboxMobileEvents/MMENSURLSessionWrapper.m
rclee: You're right it doesn't make it to the delegate method in that case. I was testing on the wrong build... I'll add `self` back as the delegate. 🤦‍♂

julianrex: Watch for the retain cycles!

andrlee: @julianrex can we catch retain cycles with static analysis tools?

alfwatt:
Any update on this fix?

rclee:
Got sidetracked but I'm hopping back in the saddle.

 **In-progess**: did some testing yesterday to verify requests are making it through and how re-initializing our MMEEventsManager using `-initializeWithAccessToken:userAgentBase:hostSDKVersion:` affects things in relation to memory leaks.

rclee:
Ready for another a review here :)

alfwatt (approved):
LGTM

MapboxMobileEvents/MMENSURLSessionWrapper.m
```diff
@@ -26,6 +34,10 @@ - (void)reconfigure:(MMEEventsConfiguration *)configuration {
     [self.certPin updateWithConfiguration:configuration];
 }
 
+-(void)dealloc {
+    [self.session invalidateAndCancel];
```
julianrex, line 38: I'm not sure this is going to fire @rclee @alfwatt: since `sessionWithConfiguration:delegate:delegateQueue:` retains self, `dealloc` won't fire. 

We'll need to call `invalidateAndCancel` somewhere earlier.

julianrex:
Just left a comment...I think you'll still leak session wrappers.

MapboxMobileEvents/MMENSURLSessionWrapper.m
alfwatt: Yup. Good catch.

rclee: So, I'm a bit concerned with calling `invalidateAndCancel` earlier or at all. It doesn't seem like we should be recreating the session at all for the life of the app ([referencing "The Eskimo"](https://forums.developer.apple.com/thread/14785)). 

We really just need to avoid having multiple `NSURLSession`/`NSURLSessionWrapper`s in play. I think that may be what's causing our troubles in the first place.

Issue #128, opened 2019-04-12
⚠️ Minor analyze warnings


julianrex:
Running `analyze` in the context of mapbox-gl-native, gives a few warnings:

- `- (NSError *)statusErrorFromRequest:(NSURLRequest *)request andHTTPResponse:` can return `nil`, but is marked as returning non-null.
- False positives of leaks in `reachabilityWithHostname:` & `reachabilityWithAddress:`
- Some warnings about localized strings (which IIRC are debug?)

Can we fix or mute these?

rclee:
Thanks for reporting these @julianrex! We should really start running these checks... which has been flagged https://github.com/mapbox/mapbox-events-ios/issues/72.

@alfwatt

julianrex:
See https://github.com/mapbox/mapbox-gl-native/pull/14572/commits/54078f68584fb3ea609962d1c0e4e51d4ac4464a - change to clean up the equivalent reachability warnings in the Maps SDK.

alfwatt:
Done!

PR #131, opened 2019-05-10
Fix minor warnings

rclee:
This should fix some warnings we were seeing when running `analyze` on Xcode. Excluding the `nullable` fix since we have that already in an upcoming PR.

ref: https://github.com/mapbox/mapbox-events-ios/issues/128


alfwatt (approved):
LGTM

PR #139, opened 2019-06-27
Allow subclassing MMEEventsManager

frederoni:
We subclass MMEEventsManager in [mapbox-navigation-ios](https://github.com/mapbox/mapbox-navigation-ios)’s test suite and use dependency injection to validate telemetry collection.

Our tests started failing on https://github.com/mapbox/mapbox-navigation-ios/pull/2151 when we bumped the events library to v0.9.4. This was caused by a newly introduced singleton return value in MMEEventsManager initializer, which seems to be a guard against subclassing or just to ensure that the singleton is the only instance of MMEEventsManager.

Let me know if you have a better idea of how we could create a mock instance of MMEEventsManager to avoid these changes.

cc @1ec5 @rclee @alfwatt

### DIFF
--- a/MapboxMobileEvents/MMEEventsManager.m
+++ b/MapboxMobileEvents/MMEEventsManager.m
@@ -53,14 +53,10 @@
     
     dispatch_once(&onceToken, ^{
         [MMECategoryLoader loadCategories];
-        _sharedManager = [MMEEventsManager.alloc initShared];
+        _sharedManager = [[self alloc] initShared];
     });
     
     return _sharedManager;
-}
-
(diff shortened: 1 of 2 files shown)


alfwatt:
@frederoni the design intent with the singleton is to prevent clients from creating multiple instances.

When we need a standalone instance for our own tests we use the `initShared` private initializer:

`MMEEventsManager *eventsManager = [MMEEventsManager.alloc initShared];`

frederoni:
Thanks for the update. We worked around it in our tests by using `initShared`.

PR #142, opened 2019-07-03
Fix Malformed Error Reports, Invalid Logging Events and Quiets Debug Logging

alfwatt:
Fixes: mapbox/mobile-telemetry#380
Fixes: mapbox/mobile-telemetry#381

### DIFF
--- a/MapboxMobileEvents/MMEEvent.m
+++ b/MapboxMobileEvents/MMEEvent.m
@@ -325,7 +325,7 @@
             if (![eventAttributes.allKeys containsObject:MMEEventKeyEvent]) { // is required
                 *error = [NSError errorWithDomain:MMEErrorDomain code:MMEErrorEventInitMissingKey userInfo:@{
                     NSLocalizedDescriptionKey: @"eventAttributes does not contain MMEEventKeyEvent",
-                    MMEErrorEventAttributesKey: eventAttributes
+                    MMEErrorEventAttributesKey: eventAttributes ?: NSNull.null
                 }];
                 self = nil;
             }
@@ -343,7 +343,7 @@
         else {
             *error = [NSError errorWithDomain:MMEErrorDomain code:MMEErrorEventInitInvalid userInfo:@{
                 NSLocalizedDescriptionKey: @"eventAttributes is not a valid JSON Object",
-                MMEErrorEventAttributesKey: eventAttributes
+                MMEErrorEventAttributesKey: eventAttributes ?: NSNull.null
             }];
             self = nil;
         }
@@ -352,7 +352,7 @@
         *error = [NSError errorWithDomain:MMEErrorDomain code:MMEErrorEventInitException userInfo:@{
             NSLocalizedDescriptionKey: @"exception processing eventAttributes",
             MMEErrorUnderlyingExceptionKey: eventAttributesException,
-            MMEErrorEventAttributesKey: eventAttributes
+            MMEErrorEventAttributesKey: eventAttributes ?: NSNull.null
         }];
         self = nil;
     }
--- a/MapboxMobileEvents/MMEEvent.m
+++ b/MapboxMobileEvents/MMEEvent.m
@@ -9,6 +9,8 @@
 #import "UIKit+MMEMobileEvents.h"
 #endif
 
+#import <CommonCrypto/CommonDigest.h>
+
 @interface MMEEvent ()
 @property(nonatomic,retain) MMEDate *dateStorage;
 @property(nonatomic,retain) NSDictionary *attributesStorage;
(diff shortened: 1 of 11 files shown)


rclee (dismissed):

rclee (dismissed):

MapboxMobileEvents/MMEEvent.m
```diff
@@ -135,6 +134,27 @@ + (instancetype)crashEventReporting:(NSError *)eventsError error:(NSError **)cre
...
+        if (callstackDigest) {
+            uint8_t digest[CC_SHA224_DIGEST_LENGTH];
+            CC_SHA224(callstackDigest.bytes, (unsigned)callstackDigest.length, digest);
+            NSMutableString* output = [NSMutableString stringWithCapacity:CC_SHA224_BLOCK_BYTES];
+            for(int i = 0; i < CC_SHA224_DIGEST_LENGTH; i++) {
+                [output appendFormat:@"%02x", digest[i]];
+            }
+
+            if (output) {
+                crashAttributes[MMEEventKeyStackTraceHash] = output;
```
julianrex, line 149: What's the hash used for? Is it just to "bucket" crashes together?

MapboxMobileEvents/MMEEvent.m
```diff
@@ -296,41 +317,41 @@ - (instancetype)init {
...
-    }
-    else if ([NSJSONSerialization isValidJSONObject:eventAttributes]) {
-        if (![eventAttributes.allKeys containsObject:MMEEventKeyEvent]) { // is required
-            *error = [NSError errorWithDomain:MMEErrorDomain code:MMEErrorEventInit userInfo:@{
-                MMEErrorDescriptionKey: @"eventAttributes does not contain MMEEventKeyEvent",
-                MMEErrorEventAttributesKey: eventAttributes
-            }];
-            self = nil;
+    @try {
+        if (eventAttributes == MMEEvent.nilAttributes) { // special case for initFromCoder
```
julianrex, line 321: can `eventAttributes` be `nil`?

MapboxMobileEvents/MMEEvent.m
```diff
@@ -296,41 +317,41 @@ - (instancetype)init {
...
+                        eventAttributesStorage[MMEEventKeyCreated] = [MMEDate.iso8601DateFormatter stringFromDate:_dateStorage];
+                    }
+
+                    self.attributesStorage = eventAttributesStorage;
+            }
+        }
+        else {
+            *error = [NSError errorWithDomain:MMEErrorDomain code:MMEErrorEventInitInvalid userInfo:@{
+                NSLocalizedDescriptionKey: @"eventAttributes is not a valid JSON Object",
+                MMEErrorEventAttributesKey: eventAttributes
```
julianrex: ... this will crash if `eventAttributes` is nil. Perhaps add `?: NSNull.null`

MapboxMobileEvents/MMEEvent.m
```diff
@@ -296,41 +317,41 @@ - (instancetype)init {
...
         }
     }
-    else {
-        *error = [NSError errorWithDomain:MMEErrorDomain code:MMEErrorEventInit userInfo:@{
-            MMEErrorDescriptionKey: @"eventAttributes is not a valid JSON Object",
+    @catch(NSException* eventAttributesException) {
+        *error = [NSError errorWithDomain:MMEErrorDomain code:MMEErrorEventInitException userInfo:@{
+            NSLocalizedDescriptionKey: @"exception processing eventAttributes",
+            MMEErrorUnderlyingExceptionKey: eventAttributesException,
             MMEErrorEventAttributesKey: eventAttributes
```
julianrex: same here

julianrex (approved):
Just the question about whether the `eventAttributes` parameter can ever be nil. But otherwise looks good.

MapboxMobileEvents/MMEEvent.m
alfwatt: Because we redact out any stack frames that don't come from our own libs or a couple of the system libs, we can't necessarily distinguish between two traces based on the stack trace we get back.

This hash is computed across the entire trace, including symbols, so that we can isolate how many crashes are happening in the same build of a particular app. The android SDK does something similar so this was added for parity.

MapboxMobileEvents/MMEEvent.m
alfwatt: The header has `NS_ASSUME_NONNULL_BEGIN` enabled, so it'll generate a warning at compile time if a real nil is provided. `MMEEvent.nilAttributes` is a special object we use to allow for constructing an event which we can then fill in with initFromCoder.

MapboxMobileEvents/MMEEvent.m
alfwatt: The caller would have to get past the `NS_ASSUME_NONNULL_BEGIN` checks, and this particular one is inside a `@try { ... }` block, but one more check can't hurt (also added on line 328)

julianrex: Answered above. Resolving.

PR #147, opened 2019-07-12
Deprecate map tap and map drag event types

friedbunny:
Deprecates `MMEEventTypeMapTap` and `MMEEventTypeMapDragEnd`, two stragglers from https://github.com/mapbox/mapbox-events-ios/pull/141.

/cc @alfwatt @rclee


alfwatt (approved):
Thanks for the typo fix

PR #154, opened 2019-07-19
[WIP] Refactor debug events 

rclee:
`MMEEventsManager.sharedManager` was not allowing for the events to be queued in the tests.

Further changes may be needed where debug events are found.


alfwatt:
Do we have a test for the old-style debug events? We should add one which tries to send one and make sure it's caught on our side.

alfwatt:
Are these changes stale?

rclee:
:nod:

Issue #155, opened 2019-08-02
🔨 Add distance filtering when updating the geofence


rclee:
**Feature**
In order to prevent excess updates to CoreLocation delegate methods, we'd like to add distance filtering based on the center of the last created geofence. 

**Customers**
Developers using MapboxMobileEvents should see a reduction in delegate calls.

**Impact**
End users may see an improvement in battery drain.

rclee:
Complete: https://github.com/mapbox/mapbox-events-ios/commit/b3f7325363eb90fff1908f1fea90de230e01e92d

PR #156, opened 2019-08-12
Add distance filtering to region monitoring

rclee:
Testing required. 

ref: https://github.com/mapbox/mapbox-events-ios/issues/155


alfwatt:
Please add:
- Unit Test Cases 
- Test Plan (probably needs a GPX file for the simulator with a bunch of points inside the area and another with points that eventually lead outside of the area)
- ~~Configurable radii for both the fence and the filter~~

alfwatt:
Stale

PR #158, opened 2019-08-16
Pending metrics error handling

rclee:
When going from event version v2 to v1 a crash occurred because `event.name` was missing on pending events. 

This PR fixes this issue by deleting the malformed event and logging the error.

### DIFF
--- a/MapboxMobileEvents/MMEConstants.h
+++ b/MapboxMobileEvents/MMEConstants.h
@@ -161,7 +161,9 @@ typedef NS_ENUM(NSInteger, MMEErrorNumber) {
     MMEErrorEventInit = 10002,
     MMEErrorEventInitMissingKey = 10003,
     MMEErrorEventInitException  = 10004,
-    MMEErrorEventInitInvalid    = 10005
+    MMEErrorEventInitInvalid    = 10005,
+    MMEErrorEventEncoding = 10006,
+    MMEErrorEventCounting = 10007
 };
 
 /*! @brief key for MMEErrorEventInit userInfo dictionary containing the attributes which failed to create the event */
--- a/MapboxMobileEvents/MMEEvent.m
+++ b/MapboxMobileEvents/MMEEvent.m
@@ -61,7 +61,7 @@
 
 + (instancetype)eventWithAttributes:(NSDictionary *)attributes {
     NSError *eventError = nil;
-    MMEEvent *newEvent = [MMEEvent eventWithAttributes:attributes error:&eventError];
+    MMEEvent *newEvent = [self eventWithAttributes:attributes error:&eventError];
     if (eventError != nil) {
         [MMEEventsManager.sharedManager reportError:eventError];
     }
@@ -70,7 +70,7 @@
 }
 
 + (instancetype)eventWithAttributes:(NSDictionary *)attributes error:(NSError **)error {
-    return [MMEEvent.alloc initWithAttributes:attributes error:error];
+    return [self.alloc initWithAttributes:attributes error:error];
 }
 
 #pragma mark - Custom Events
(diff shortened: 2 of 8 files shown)


alfwatt:
Please rebase

MapboxMobileEvents/MMEConstants.h
```diff
@@ -161,7 +161,8 @@ typedef NS_ENUM(NSInteger, MMEErrorNumber) {
     MMEErrorEventInit = 10002,
     MMEErrorEventInitMissingKey = 10003,
     MMEErrorEventInitException  = 10004,
-    MMEErrorEventInitInvalid    = 10005
+    MMEErrorEventInitInvalid    = 10005,
+    MMEErrorEncoding = 10006
```
alfwatt: Can we make this `MMErrorEventEncoding`

MapboxMobileEvents/MMEEvent.m
```diff
@@ -4,6 +4,7 @@
 #import "MMECommonEventData.h"
 #import "MMEReachability.h"
 #import "MMEEventsManager.h"
+#import "MMEEventLogger.h"
```
alfwatt: Is this still needed?

MapboxMobileEvents/MMEEvent.m
```diff
@@ -431,12 +432,17 @@ - (id)copyWithZone:(NSZone *)zone {
...
     if (self = [self init]) {
         NSInteger encodedVersion = [aDecoder decodeIntegerForKey:MMEEventVersionKey];
-        _dateStorage = [aDecoder decodeObjectOfClass:MMEDate.class forKey:MMEEventDateKey];
-        _attributesStorage = [aDecoder decodeObjectOfClass:NSDictionary.class forKey:MMEEventAttributesKey];
+        
         if (encodedVersion > MMEEventVersion2) {
-            NSLog(@"%@ WARNING encodedVersion %li > MMEEventVersion %li",
-                NSStringFromClass(self.class), (long)encodedVersion, (long)MMEEventVersion1);
+            NSString *errorString = [[NSString alloc] initWithFormat:@"%@ %@ encodedVersion %li > MMEEventVersion %li",
+                                     self.name, NSStringFromClass(self.class), (long)encodedVersion, (long)MMEEventVersion1];
```
alfwatt: self.name isn't supposed to ever be nil, but just in case let's use `self.name ?: @"nil"`

MapboxMobileEvents/MMEMetricsManager.m
```diff
@@ -126,9 +126,11 @@ - (void)updateMetricsFromEventQueue:(NSArray *)eventQueue {
         self.metrics.eventCountTotal += (int)eventQueue.count;
         
         for (MMEEvent *event in eventQueue) {
-            NSNumber *eventCount = [self.metrics.eventCountPerType objectForKey:event.name];
-            eventCount = [NSNumber numberWithInteger:[eventCount integerValue] + 1];
-            [self.metrics.eventCountPerType setObject:eventCount forKey:event.name];
+            if (event.name) {
```
alfwatt, line 129: An unnamed event at this point in execution is a reportable error, please add an else block with an error report and a new code for tracking

alfwatt (changes requested):
Please see notes and squash into a single commit

MapboxMobileEvents/MMEMetricsManager.m
rclee: I was considering removing this check but it's probably good to keep in place. Error report added. Thanks

MapboxMobileEvents/MMEEvent.m
rclee: Good catch

MapboxMobileEvents/MMEEvent.m
rclee: Nope

MapboxMobileEventsTests/MMEEventTests.mm
```diff
@@ -63,6 +65,25 @@ + (NSDictionary *)nilAttributes;
             unarchived should_not be_nil;
             unarchived should equal(event);
         });
+        
+        
+        it(@"should delete stored event when encoded version is greater than MMEEventVersion", ^{
```
alfwatt: this test description isn't quite correct, the event is encoded into memory and we validate that it decodes to nil.

we need a test to cover writing that data to the `pendingMetricsEventPath` then verify that `loadPendingTelemetryMetricsEvent` returns nil and that no file exists at the pending event path

alfwatt (approved):
Approved pending additional test case.

Please squash these 6 commits (+ an additional ones for the new test case) using [git rebase](https://blog.carbonfive.com/2017/08/28/always-squash-and-rebase-your-git-commits/)

MapboxMobileEventsTests/MMEEventTests.mm
rclee: Thanks for pointing that out. I've added the test and corrected the description.

PR #161, opened 2019-09-12
Fix compilation error on case-sensitive file system

fishred:
Fix the typo of "L" letter of the CoreLocation header filename. This is causing building
failure on OS X with file system which is case-sensitive.


alfwatt:
Good catch!

PR #163, opened 2019-10-14
Implement hacks to disable telemetry

Marcelhozeman:


alfwatt:
@Marcelhozeman N.B. The use of Mapbox SDKs and APIs on mobile devices are governed by our [Terms of Service](https://www.mapbox.com/legal/tos#[MomMom]) which require location collection be enabled, except when the end users opts-out, which they must always be able to do.

If your application  requires different terms, please contact [Mapbox Sales](https://www.mapbox.com/contact/sales/).

PR #167, opened 2019-10-23
Update SSL hashes w/ config

rclee:
This should make use of a flag in `mme_configuration` to update our SSLPinSet every time we detect a potential change in the CRL.

### DIFF
--- a/MapboxMobileEvents/Categories/NSUserDefaults+MMEConfiguration.h
+++ b/MapboxMobileEvents/Categories/NSUserDefaults+MMEConfiguration.h
@@ -4,12 +4,6 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-// MARK: mme_certificatePinningConfig Keys
-
-extern NSString * const MMEPinnedDomains;
-extern NSString * const MMEPublicKeyHashes;
-extern NSString * const MMEExcludeSubdomainFromParentPolicy;
-
 // MARK: -
 
 @class MMEDate;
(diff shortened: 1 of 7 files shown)


rclee:
Testing required

alfwatt (commented):
Looks like your brought udpateSSLPin set back, which will work in the short term, but we want to fully encapsulate the config process in `mme_configuration`:

- `mme_shouldUpdateSSL` is redundant to `mme_certificateRevocationList.length > 0` so I don't think we need it
- `updateSSLPingSet` removes the revoked hashes from `NSUserDefaults.mme_configuration.mme_certificatePinningConfig` which should be removed here: https://github.com/mapbox/mapbox-events-ios/blob/a24baa4e16a35da0d7acb121dfd5a74849192bb0/MapboxMobileEvents/Categories/NSUserDefaults%2BMMEConfiguration.m#L488

alfwatt (changes requested):
I think we should include the validation logic in `mme_updateFromConfigServiceObject ` https://github.com/mapbox/mapbox-events-ios/blob/a24baa4e16a35da0d7acb121dfd5a74849192bb0/MapboxMobileEvents/Categories/NSUserDefaults%2BMMEConfiguration.m#L541

i.e. validate the config when you read it in, and don't store invalid values I the prefs then check later.

rclee:
> Looks like your brought updateSSLPin set back, which will work in the short term, but we want to fully encapsulate the config process in mme_configuration:

> * `mme_shouldUpdateSSL` is redundant to `mme_certificateRevocationList.length > 0` so I don't think we need it

This was my initial thought as well but `mme_certificateRevocationList.length > 0` but with the current implementation we may have to update more often. If CRL > 0 then it would remain that way and each time `handleChallenge:` is called, we would be calling `updateSSLPinSet`. This is really just to fit `updateSLLPinSet` back in as a temporary fix; until we have more time to test config.

> updateSSLPingSet removes the revoked hashes from NSUserDefaults.mme_configuration.mme_certificatePinningConfig which should be removed here:

Understood, as you said: short term.

alfwatt (approved):
Looks like all the changes we paired on and the wrap-up are in good shape, thanks for the quick turnaround!

fabian-guerra (approved):
From a language perspective this looks good to me. I am not familiar with all the encoding/decoding required for this library to work. 
My only comment would be if there is a way to refactor the code to avoid using `goto`?

alfwatt:
@rclee  can you make the constants like `kMMEBackgroundGeofence` more like `MMEBackgroundGeofenceDefault`: Hungarian Notation Considered Harmfull™

PR #168, opened 2019-10-23
🐞 [iOS] mme_configuration is not applied to `isEnabled` logic in MMEEventsManager #463

alfwatt:
🐞 [iOS] mme_configuration is not applied to `isEnabled` logic in MMEE…
…ventsManager #463

- Implement `mme_isCollectionEnabled` logic
- Fix `mme_deleteObjectPersistantKey` signature
- Use the Reformed UserAgent string for error reports
- Remove config properties from `MMEEventsManager`
- Enable collection in simulator for test host
- Release the retain attributes on properties
- Simplify certificiate pinning hash matching
- Fix and Improve Tests Daily


MapboxMobileEvents/Categories/NSUserDefaults+MMEConfiguration.m
```diff
@@ -312,7 +319,21 @@ - (NSString *)mme_legacyUserAgentString {
 // MARK: - Location Collection
 
 - (BOOL)mme_isCollectionEnabled {
-    return ![self boolForKey:MMECollectionDisabled];
+    BOOL collectionEnabled = ![self boolForKey:MMECollectionDisabled];
+    
+#if TARGET_OS_SIMULATOR
+    // disable collection in the smulator unless explicitly enabled for testing
```
rclee: nit: smulator

MapboxMobileEvents/Categories/NSUserDefaults+MMEConfiguration.m
```diff
@@ -312,7 +318,21 @@ - (NSString *)mme_legacyUserAgentString {
...
+    if (!self.mme_isCollectionEnabledInSimulator) {
+        collectionEnabled = NO;
+    }
+#endif
+
+    // if not explicitly disalbed, or in simulator, check for low power mode
+    if (collectionEnabled && [NSProcessInfo instancesRespondToSelector:@selector(isLowPowerModeEnabled)]) {
+        collectionEnabled = !NSProcessInfo.processInfo.isLowPowerModeEnabled;
```
rclee: We may need to bring back some of the previous code to keep some warnings silent for the Maps folks. iOS 8 is still supported by us (barely), we may need to keep in a OS check as well.
```
// Only check power mode if compiling with the iOS 9+ SDK and, at runtime, if the API exists
#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 90000
    if ([NSProcessInfo instancesRespondToSelector:@selector(isLowPowerModeEnabled)]) {
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wpartial-availability"
        isPowerModeCompatibleWithCollection = ![[NSProcessInfo processInfo] isLowPowerModeEnabled];
#pragma clang diagnostic pop
    }
#endif
```

alfwatt: I was expecting to get a warning on that, and double checked the targets set setup for iOS 8 as the target API. I think the `instancesRespondToSelector:` check may be sufficient to squash the warning. I would setup a simulator to check, but the oldest one available is iOS 10!

MapboxMobileEvents/MMEEventsManager.h
```diff
@@ -16,40 +16,7 @@ NS_ASSUME_NONNULL_BEGIN
...
-/*! @brief Mapbox Access Token
-    @link https://account.mapbox.com */
-@property (nonatomic, copy) NSString *accessToken;
-
-/*! @brief baseURL */
-@property (nonatomic, null_resettable) NSURL *baseURL;
-
-/*! @brief accountType */
-@property (nonatomic) NSInteger accountType;
-
```
rclee, line 52: I'm a little concerned that some of these public properties may be used in other frameworks. Should we consider deprecating?

rclee (approved):
I left a final note to consider but everything else LGTM!

(enjoy this gif from Independence Day)

MapboxMobileEvents/MMEEventsManager.h
rclee: I meant to highlight most of the properties in `MMEEventsManager`... that didn't work.

MapboxMobileEvents/Categories/NSUserDefaults+MMEConfiguration.m
alfwatt: Whoops, forgot: CLANG_WARN_UNGUARDED_AVAILABILITY = YES_AGGRESSIVE

MapboxMobileEvents/MMEEventsManager.h
alfwatt: The nice thing about `0.x.x` releases, is that we don't promise API stability at all…

PR #178, opened 2019-11-13
Tweaks to support integration with mapbox-gl-native-ios

julianrex:
A few changes related to mapbox-gl-native-ios's integration:

- Adds `NSUserDefaults+MMEConfiguration.h` to public headers & include copy phase
- Adds a `RelWithDebInfo` configuration (currently just a duplicate of `Release` but with debug symbols) - this is to match gl-native-ios, which in turns matches gl-native. It's not the end of the world if we want to skip this change, but it will require defining an explicit dependency rather than relying on Xcode.
- Added a script (`create_bundled_framework.sh`) that creates a `modulemap` file that's used as a submodule in `Mapbox.framework` (when events is linked statically) - this allows the usage of `@import MapboxMobileEvents` (though does require `@import Mapbox`). Related to this, the headers in `MapboxMobileEvents.h` have changed to use `""` rather than `<>`; without this we end up with a couple of different warnings at the application level.


rclee (approved):
LGTM

PR #187, opened 2019-12-11
Log Debug Events

rclee:
`MMEEventsManager.sharedManager` was not allowing for the events to be queued in the tests.

`debugEvent`s shouldn't be pushed, only logged.


alfwatt (approved):
Better API usage overall, LGTM

PR #189, opened 2019-12-19
Add test coverage

rclee:


rclee:
Majority of this is done if you have some time to look it over @alfwatt 👀

rclee:
`MMECertPin` and `MMEDate` has a few more things that could be tested. After we remove the debug classes (that don't need tests here) we should be over 90% on code coverage.

MapboxMobileEvents/MMEMetricsManager.m
```diff
@@ -214,7 +198,7 @@ - (void)updateConfigurationJSON:(NSDictionary *)configuration {
 }
 
 - (void)updateCoordinate:(CLLocationCoordinate2D)coordinate {
-    if (!self.metrics.deviceLat || !self.metrics.deviceLon) {
+    if (!self.metrics.deviceLat && !self.metrics.deviceLon) {
```
alfwatt, line 211: Yikes

Tests/NSUserDefaults+MMEConfigurationTests.m
```diff
@@ -87,6 +87,77 @@ - (void)testConfigEventTagDefault {
...
+        @"MMEMapboxHostSDKVersion": @"1.0.0"
+    }];
+    [NSUserDefaults.mme_configuration mme_registerDefaults];
+
+    XCTAssertFalse(NSUserDefaults.mme_configuration.mme_isCollectionEnabled);
+    XCTAssert([NSUserDefaults.mme_configuration.mme_legacyUserAgentBase isEqualToString:@"com.mapbox.test"]);
+    XCTAssert([NSUserDefaults.mme_configuration.mme_legacyHostSDKVersion isEqualToString:@"1.0.0"]);
+}
+
+//TODO: Persistent object deletion isn't working as expected
```
alfwatt: We should resolve this before merging these changes

MapboxMobileEvents/Categories/NSString+MMEVersions.m
```diff
@@ -106,44 +106,45 @@ - (NSInteger)mme_semverPatchVersion {
...
-    if (regexMatch && regexMatch.numberOfRanges >= MMESemverReleaseIndex) {
-        buildString = [self substringWithRange:[regexMatch rangeAtIndex:MMESemverReleaseIndex]];
-    }
-    return buildString;
-}
-
-- (NSArray<NSString *>*)mme_semverPreReleaseIdentifiers {
-    return [[self mme_semverPreReleaseComponent] componentsSeparatedByString:@"."];
-}
+//TODO: These methods are not working as expected in tests and are not used currently
```
alfwatt: IIRC there aren't tests for these, we should add some if we're going to keep the code at all (given it's not currently used, it could be deleted from the project until it's needed)

alfwatt (commented):
Couple of smaller issues

MapboxMobileEvents/MMEMetricsManager.m
rclee: Yup

MapboxMobileEvents/Categories/NSString+MMEVersions.m
rclee: When writing a couple of the tests it wasn't behaving as I would have expected. Removing the unused code for now seems like a good bet.

Tests/NSUserDefaults+MMEConfigurationTests.m
alfwatt: @rclee looking at testDeletePersistentObject we probably shoudn’t be mixing that test with MMECollectionDisabled which has some special cases

I suggest re-writing the test to use mme_setObject:forPersistentKey: (which you’ll need to add to the private header, maybe along with it’s volatile counterpart

rclee: Done, thanks!

alfwatt (approved):
LGTM

PR #190, opened 2019-12-21
Refactor debug process

rclee:
Some code to move and remove. This should bring down the size of our binary.


alfwatt (commented):
Is there a corresponding change in the example app to keep these reports or are we just removing them?

rclee:
Example app is next. Working on adding in the hook for that to work now.

alfwatt (approved):
#ifdef  CODE_WORKING
GitHubCommit()
#endif

PR #192, opened 2020-01-15
Fix trailing semicolon warning

keith:
```
MapboxMobileEvents/UIKit+MMEMobileEvents.m:5:33: error: extra ';' outside of a function [-Werror,-Wextra-semi]
void mme_linkUIKitCategories(){};
                                ^
```

### DIFF
--- a/MapboxMobileEvents/Categories/UIKit+MMEMobileEvents.m
+++ b/MapboxMobileEvents/Categories/UIKit+MMEMobileEvents.m
@@ -2,7 +2,7 @@
 
 @implementation NSObject (MMEMobileEvents)
 
-void mme_linkUIKitCategories(){};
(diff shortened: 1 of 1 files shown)


bsudekum:
cc @rclee @1ec5

1ec5 (approved):
Looks good to me. @rclee, not sure if our existing build systems are set up well to detect problems like this. Someone integrating MapboxMobileEvents into their application using CocoaPods is going to see this warning, but neither the gl-native-ios project nor the navigation SDK repository are set up to build MapboxMobileEvents the way CocoaPods does. (We do have a CocoaPods integration test in the navigation SDK repository; I guess we could have it fail on warnings too…)

rclee (approved):

keith:
Thanks!

PR #195, opened 2020-02-04
Add debug logging enablement to bundle

rclee:
Adds a plist enablement of debug logging on application level instead of host SDK.


alfwatt (approved):
Approved pending a test plan in TESTING.md (which we should create) and add documentation for how to enable this logging in the README.md

PR #203, opened 2020-03-25
Add 13.4 location attributes (speed and course accuracy)

rclee:
Adding new attributes to track as part of our location probes to enhance traffic related products.

Testing notice: these attributes require a minimum version of iOS 13.4 and Xcode 11.4.


rclee:
Looks like the tests are failing because CI's Xcode version has yet to update...

misterkerry (approved):
lgtm.  Question @rclee how are the requirements (per iOS 13.4 for instance) propagated?  Is there an update to the SDK's README.md?

MapboxMobileEvents/Categories/CLLocation+MMEMobileEvents.m
```diff
@@ -24,6 +24,20 @@ - (CLLocationAccuracy)mme_roundedVerticalAccuracy {
...
 }
 
+- (CLLocationAccuracy)mme_roundedSpeedAccuracy {
+    if (@available(iOS 13.4, *)) {
+        return round(self.speedAccuracy);
+    }
+    return 0;
+}
+
+- (CLLocationAccuracy)mme_roundedCourseAccuracy {
```
dmiluski, line 31: https://github.com/mapbox/mapbox-events-ios/pull/203/files#diff-bd4c68f875c0aa434bb834c6903bac29R15

I'm wondering if there is precedent for when to expose functions? Vs when to default them?

I see this CoreLocation method isn't usable if not iOS 13.4+? I understand 0 accuracy means it's not accurate. But I also wonder if the existence (not available vs 0) would lead to more confusion?
```
@available(iOS 13.4, *)
- (CLLocationAccuracy)mme_roundedSpeedAccuracy
```

rclee: This is a good point. Since this is not a required attribute, marking availability makes more sense.

rclee:
Marked availability on these internal APIs. Holding off on merge until we get the tests passing...

Issue #47, opened 2018-06-25
Handle 'sharedApplication' is unavailable: not available on iOS (App Extension)


bsudekum:
I'm getting an error when using this library inside of an extension:

```
'sharedApplication' is unavailable: not available on iOS (App Extension) - Use view controller based solutions where appropriate instead.
```

This is a pretty widespread and documented issue, [ref](https://miqu.me/blog/2016/11/28/app-extensions-xcode-and-cocoapods-omg/).

/cc @rclee

friedbunny:
This could be addressed alongside #44, as it’s more or less the same codepath.

dmiluski:
Fix merged, should be available in the next release

PR #204, opened 2020-03-30
iOS (App Extension): Friendly Support

dmiluski:
| Obligatory Gif |
| - |
| tenor |
| Extension |

## 🔧  Issue:
https://github.com/mapbox/mapbox-events-ios/issues/47#event-3180214494

## 🥅 Goal:
Play friendly with usage inside of an extension

## 📚  Context:
When using this module inside of an extension, developers will see errors in console:
'sharedApplication' is unavailable: not available on iOS (App Extension) - Use view controller based solutions where appropriate instead.

## 🗺 Approach:
1. Leverage view controller based solutions where appropriate (Available in iOS10 +
2. Inspect bundle to determine if if an extension or not to determine whether or not to access model information with Application or ExtensionContext

## 👷 Changes:
+ Add NSBundle Extensin to inspect whether in an Extension
+ Add NSExtensionContext to extract contentSizeScale from view controller based solution.
+ Inspect Bundle/Extension to construct Scale/State values

## 🔨 Unrelated Change:
+ Included DS_Store so it doesn’t clutter git diff

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,3 +68,4 @@ fastlane/test_output
 
 test_output/
 docs/
+.DS_Store
--- a/MapboxMobileEvents/Categories/UIKit+MMEMobileEvents.m
+++ b/MapboxMobileEvents/Categories/UIKit+MMEMobileEvents.m
@@ -82,3 +82,46 @@ void mme_linkUIKitCategories(){}
 }
 
 @end
+
+@implementation NSExtensionContext (MMEMobileEvents)
+
++ (NSInteger)mme_contentSizeScale {
+
+    NSInteger result = -9999;
+
+    if (@available(iOS 10, *)) {
+        NSString *sc = UIScreen.mainScreen.traitCollection.preferredContentSizeCategory;
+
+        if ([sc isEqualToString:UIContentSizeCategoryExtraSmall]) {
+            result = -3;
+        } else if ([sc isEqualToString:UIContentSizeCategorySmall]) {
+            result = -2;
+        } else if ([sc isEqualToString:UIContentSizeCategoryMedium]) {
+            result = -1;
+        } else if ([sc isEqualToString:UIContentSizeCategoryLarge]) {
+            result = 0;
+        } else if ([sc isEqualToString:UIContentSizeCategoryExtraLarge]) {
+            result = 1;
+        } else if ([sc isEqualToString:UIContentSizeCategoryExtraExtraLarge]) {
+            result = 2;
+        } else if ([sc isEqualToString:UIContentSizeCategoryExtraExtraExtraLarge]) {
+            result = 3;
+        } else if ([sc isEqualToString:UIContentSizeCategoryAccessibilityMedium]) {
+            result = -11;
+        } else if ([sc isEqualToString:UIContentSizeCategoryAccessibilityLarge]) {
+            result = 10;
+        } else if ([sc isEqualToString:UIContentSizeCategoryAccessibilityExtraLarge]) {
+            result = 11;
+        } else if ([sc isEqualToString:UIContentSizeCategoryAccessibilityExtraExtraLarge]) {
+            result = 12;
+        } else if ([sc isEqualToString:UIContentSizeCategoryAccessibilityExtraExtraExtraLarge]) {
+            result = 13;
+        }
+    } else {
+        // No-Op
+    }
+
+    return result;
+}
+
+@end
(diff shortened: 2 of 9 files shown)


.gitignore
```diff
@@ -68,3 +68,4 @@ fastlane/test_output
 
 test_output/
 docs/
+.DS_Store
```
dmiluski, line 71: Added DS_Store ignore to limit git diff clutter

MapboxMobileEvents/Categories/UIKit+MMEMobileEvents.m
```diff
@@ -82,3 +82,46 @@ - (NSInteger)mme_contentSizeScale {
...
 @end
+
+@implementation NSExtensionContext (MMEMobileEvents)
+
++ (NSInteger)mme_contentSizeScale {
+
+    NSInteger result = -9999;
+
+    if (@available(iOS 10, *)) {
+        NSString *sc = UIScreen.mainScreen.traitCollection.preferredContentSizeCategory;
```
dmiluski, line 93: Given UIApplication shared isn't available, grab preferredContentSizeCategory directly from UIScreen. Unfortunately this only became available in iOS10. Based on prior result of -9999 value, assumed this same default would be used in the event user is running < iOS10

MapboxMobileEvents/MMECommonEventData.m
```diff
@@ -71,16 +72,22 @@ - (instancetype)init {
...
-            return MMEApplicationStateForeground;
-        case UIApplicationStateInactive:
-            return MMEApplicationStateInactive;
-        case UIApplicationStateBackground:
-            return MMEApplicationStateBackground;
-        default:
-            return MMEApplicationStateUnknown;
+
+    if (NSBundle.mme_isExtension) {
+        return MMEApplicationStateUnknown;
```
dmiluski: I wasn't quite sure how to port these values in the event of running in an extension. Extensions offer a much more limiting running time being killed immediately after dismissal.

Options:
- Return active by default if in extension?
- Return Unknown given there isn't a value to check?
- Other?

MapboxMobileEvents/MMEEvent.m
```diff
@@ -231,7 +232,12 @@ + (instancetype)mapLoadEventWithDateString:(NSString *)dateString commonEventDat
     eventAttributes[MMEEventKeyOperatingSystem] = commonEventData.osVersion;
     eventAttributes[MMEEventKeyResolution] = @(commonEventData.scale);
 #if TARGET_OS_IOS || TARGET_OS_TVOS
-    eventAttributes[MMEEventKeyAccessibilityFontScale] = @(UIApplication.sharedApplication.mme_contentSizeScale);
+
+    if (NSBundle.mme_isExtension) {
```
dmiluski, line 236: Differentiate value sourcing based on whether extension or app.

misterkerry (approved):
lgtm

MapboxMobileEvents/MMECommonEventData.m
dmiluski: Alternative:
Drop key/value all together if extension? Or the full event if extension?

rclee: Let's add another value for `MMEApplicationStateExtension` and merge. We discussed this in a recent meeting and think it makes sense to let this fly with a known value to observe.

rclee (approved):
Approved pending addition to ApplicationState. 

As an aside, how come we chose to fork here? I think Bitrise is having some trouble figuring out how to handle that...

dmiluski:
> Approved pending addition to ApplicationState.
> 
> As an aside, how come we chose to fork here? I think Bitrise is having some trouble figuring out how to handle that...

@rclee . Out of habit, before contributing to any repo I usually fork if option is available. This allows for contribution with or without collaborator status. Given this is a public repo, and others could also potentially contribute, this seemed okay at least for the time being? (I'll have to take a peak at bitrise and see if fork CI support is allowed)

A fork allows for a scratch pad to follow any organizational branch naming convention given any change I make doesn't clutter the main repo. In the past, if I > 10 branches I have felt that there may be a lack of focus on delivering features/value. I have seen 1,000+ branches in a repo and its usually a sign that team members aren't delivering features, or perhaps even worse, aren't collaborating to push work to the finish line and letting branches linger.

In doing so the main forked repo is allowed to remain focused and easy to discover what features are in the pipeline with long running feature branches such as Swift 3/4/5 support able to be done with focus, while features/bugfixes can be approached in the confines of a PR.

dmiluski:
@rclee https://github.com/mapbox/mapbox-events-ios/pull/204#discussion_r409245249

Will do

dmiluski:
There appears to be an issue with PRs from Forks. While investigating, will re-open from a local main repo branch.

PR #205, opened 2020-04-16
iOS (App Extension): Friendly Support

dmiluski:
| Obligatory Gif |
| - |
| tenor |
| Extension |

## 🔧  Issue:
https://github.com/mapbox/mapbox-events-ios/issues/47#event-3180214494

## 🥅 Goal:
Play friendly with usage inside of an extension

## 📚  Context:
When using this module inside of an extension, developers will see errors in console:
'sharedApplication' is unavailable: not available on iOS (App Extension) - Use view controller based solutions where appropriate instead.

## 🗺 Approach:
1. Leverage view controller based solutions where appropriate (Available in iOS10 +
2. Inspect bundle to determine if if an extension or not to determine whether or not to access model information with Application or ExtensionContext

## 👷 Changes:
+ Add NSBundle Extensin to inspect whether in an Extension
+ Add NSExtensionContext to extract contentSizeScale from view controller based solution.
+ Inspect Bundle/Extension to construct Scale/State values

## 🔨 Unrelated Change:
+ Included DS_Store so it doesn’t clutter git diff

### DIFF
--- a/MapboxMobileEvents/MMEDate.m
+++ b/MapboxMobileEvents/MMEDate.m
@@ -55,6 +55,8 @@ static NSTimeInterval _timeOffsetFromServer = 0.0; // TODO maintain a list of MM
     if (!_httpDateFormatter) {
         _httpDateFormatter = [[NSDateFormatter alloc] init];
         [_httpDateFormatter setDateFormat:@"E, d MMM yyyy HH:mm:ss Z"];
+        [_httpDateFormatter setTimeZone:[NSTimeZone timeZoneWithName:@"UTC"]];
+
     }
 
     return _httpDateFormatter;
--- a/Tests/MMEDateTests.m
+++ b/Tests/MMEDateTests.m
@@ -15,8 +15,7 @@
 - (void)testHTTPDateFormatter {
     NSDate *date = [NSDate dateWithTimeIntervalSince1970:214123];
     NSString *dateString = [MMEDate.HTTPDateFormatter stringFromDate:date];
-    
-    XCTAssert([dateString isEqualToString:@"Sat, 3 Jan 1970 03:28:43 -0800"]);
+    XCTAssert([dateString isEqualToString:@"Sat, 3 Jan 1970 11:28:43 +0000"], @"%@", dateString);
 }
 
 - (void)testNSSecureCoding {
--- a/.gitignore
+++ b/.gitignore
@@ -68,3 +68,4 @@ fastlane/test_output
 
 test_output/
 docs/
+.DS_Store
--- a/MapboxMobileEvents.xcodeproj/project.pbxproj
+++ b/MapboxMobileEvents.xcodeproj/project.pbxproj
@@ -203,6 +203,9 @@
 		CF93A6582241EDD500BBA199 /* MMECertPin.m in Sources */ = {isa = PBXBuildFile; fileRef = CF93A6562241EDD500BBA199 /* MMECertPin.m */; };
 		CF93A6592241EDD500BBA199 /* MMECertPin.m in Sources */ = {isa = PBXBuildFile; fileRef = CF93A6562241EDD500BBA199 /* MMECertPin.m */; platformFilter = ios; };
 		CF93A65B2241EDD500BBA199 /* MMECertPin.h in Headers */ = {isa = PBXBuildFile; fileRef = CF93A6572241EDD500BBA199 /* MMECertPin.h */; };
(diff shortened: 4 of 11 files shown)


dmiluski:
Filing from main repo instead of fork given Bitrise trouble with fork/PR:ID
https://github.com/mapbox/mapbox-events-ios/pull/204

MapboxMobileEvents/MMEDate.m
```diff
@@ -55,6 +55,7 @@ + (NSDateFormatter *)HTTPDateFormatter {
     if (!_httpDateFormatter) {
         _httpDateFormatter = [[NSDateFormatter alloc] init];
         [_httpDateFormatter setDateFormat:@"E, d MMM yyyy HH:mm:ss Z"];
+        [_httpDateFormatter setTimeZone:[NSTimeZone timeZoneWithName:@"PST"]];
```
dmiluski: @alfwatt @rclee ,

One of the errors I got once the CI was up and running again was due to time zone difference:
```
MMEDateTests
  testHTTPDateFormatter, (([dateString isEqualToString:@"Sat, 3 Jan 1970 03:28:43 -0800"]) is true) failed - Sat, 3 Jan 1970 11:28:43 +0000
  /Users/vagrant/git/Tests/MMEDateTests.m:18
    NSString *dateString = [MMEDate.HTTPDateFormatter stringFromDate:date];
    XCTAssert([dateString isEqualToString:@"Sat, 3 Jan 1970 03:28:43 -0800"], @"%@", dateString);
}
```

For an experiment, I hard coded PST here to toy around double checking the expectation validation.

In this case, the test is expecting the string to include a time zone `Z` of PST. I'm guessing the CI box is somewhere in the UK UTC timezone as bitrise scales which may have revealed this issue. 

I'm looking to see what the expectation was for this behavior? Does this need the time zone Z? If not, we could send a timestamp formatted to a hardcoded UTC which would allow us to ensure the test passes regardless of which timezone the CI box is run.

Tests/MMEDateTests.m
```diff
@@ -15,8 +15,7 @@ - (void)setUp {
 - (void)testHTTPDateFormatter {
     NSDate *date = [NSDate dateWithTimeIntervalSince1970:214123];
     NSString *dateString = [MMEDate.HTTPDateFormatter stringFromDate:date];
-    
-    XCTAssert([dateString isEqualToString:@"Sat, 3 Jan 1970 03:28:43 -0800"]);
+    XCTAssert([dateString isEqualToString:@"Sat, 3 Jan 1970 03:28:43 -0800"], @"%@", dateString);
```
dmiluski: Added a StringFormat comment which includes a formatted comment to inspect the CI box failure which provided the necessary information to discover the timezone was off

MapboxMobileEvents/MMEDate.m
rclee: Good find! This should be a UTC check for sure. We’re only verifying formatting here.

MapboxMobileEvents/MMEDate.m
```diff
@@ -55,6 +55,8 @@ + (NSDateFormatter *)HTTPDateFormatter {
     if (!_httpDateFormatter) {
         _httpDateFormatter = [[NSDateFormatter alloc] init];
         [_httpDateFormatter setDateFormat:@"E, d MMM yyyy HH:mm:ss Z"];
+        [_httpDateFormatter setTimeZone:[NSTimeZone timeZoneWithName:@"UTC"]];
```
dmiluski, line 58: @rclee , Updated this formatter to use UTC by default. This means that the Z value will always be +0000 and also allow for test to be consistent.

Tests/MMEDateTests.m
```diff
@@ -15,8 +15,7 @@ - (void)setUp {
 - (void)testHTTPDateFormatter {
     NSDate *date = [NSDate dateWithTimeIntervalSince1970:214123];
     NSString *dateString = [MMEDate.HTTPDateFormatter stringFromDate:date];
-    
-    XCTAssert([dateString isEqualToString:@"Sat, 3 Jan 1970 03:28:43 -0800"]);
+    XCTAssert([dateString isEqualToString:@"Sat, 3 Jan 1970 11:28:43 +0000"], @"%@", dateString);
```
dmiluski, line 18: @rclee , given the move to timezone UTC, this value is now +0000 (UTC) vs -0800 (PST).

rclee (approved):
LGTM; thank you Dane!

rclee:
@dmiluski Feel free to squash and merge when ready.

PR #202, opened 2020-03-05
Mapbox Mobile Events for iOS 1.0.0-beta

alfwatt:

PR #206, opened 2020-05-05
[Test Fixes] Iterating on Failing Test Cases on Beta Branch

dmiluski:
## 🥅 Goal
Bring Beta Branch towards all passing tests

## 🔧 Issue:
https://app.zenhub.com/workspaces/mobile-telemetry-5bda50ac4b5806bc2befdbdc/issues/mapbox/mobile-telemetry/655

## 🗺️ Approach
Working my way through the build iterating over each test case

## 👨‍⚕️ Test Fixes
- [x] testAttributesDoesntExcludesNullFloat - Rounding Issue
- [x] testNSSecureCodingOfMMEEvent - Just needed a different equality check since it's an object
- [x] testTimerReachedWithEventQueued - Regression which appeared after timerManager was removed
- [x]  testRegionDistanceFilter [Invalid Test]. Unfortunately this is an indeterminate test due to async trigger of startMonitoring regions vs the array.

## 🧰 Helpers
- [x] Test Bundle Packaging to include resources in Test Target. Only need to include resources in one target now.

## 🏋️ Expectations
These particular tests should now pass.

### DIFF
--- a/MapboxMobileEvents/Categories/NSUserDefaults+MMEConfiguration_Private.h
+++ b/MapboxMobileEvents/Categories/NSUserDefaults+MMEConfiguration_Private.h
@@ -99,7 +99,7 @@ static NSTimeInterval const MMEStartupDelayMaximum = 100;
 // MARK: -
 
 @interface NSUserDefaults (MMEConfiguration_Private)
-@property(nonatomic,readonly) NSString *mme_clietnId;
+@property(nonatomic,readonly) NSString *mme_clientId;
 
 + (void)mme_resetConfiguration;
 
--- a/Tests/Categories/NSBundle+TestBundle.h
+++ b/Tests/Categories/NSBundle+TestBundle.h
@@ -1,0 +1,12 @@
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface NSBundle (TestBundle)
+
+/// Bundle containing Test Resources
++ (NSBundle *)testBundle;
+
+@end
+
+NS_ASSUME_NONNULL_END
--- a/Tests/MMEEventTests.m
+++ b/Tests/MMEEventTests.m
@@ -66,7 +66,7 @@
      MMEEvent *unarchived = [unarchiver decodeObjectOfClass:MMEEvent.class forKey:NSKeyedArchiveRootObjectKey];
 
      XCTAssertNotNil(unarchived);
-     XCTAssert(unarchived == event);
+     XCTAssertEqualObjects(unarchived, event);
 }
 
 - (void)testNSSecuredDecodeingOfMMEEvent {
--- a/Tests/MMEEventsManagerTests.m
+++ b/Tests/MMEEventsManagerTests.m
@@ -164,8 +164,6 @@
     
     [self.eventsManager pushEvent:locationEvent];
     XCTAssert(self.eventsManager.eventQueue.count > 0);
-    
-    XCTAssert(self.eventsManager.eventQueue.count == 0);
 }
 
 - (void)testCollectionNotEnabledWhilePausedAndAlwaysAuth {
--- a/Tests/MMELocationManagerTests.m
+++ b/Tests/MMELocationManagerTests.m
@@ -25,7 +25,8 @@
     
     self.locationManager = [[CLLocationManager alloc] init];
     self.mme_locationManager = [[MMELocationManager alloc] init];
-    
+
+
     self.mme_locationManager.locationManager = self.locationManager;
     self.mme_locationManager.delegate = self;
 }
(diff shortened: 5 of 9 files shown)


Tests/MMEEventTests.m
```diff
@@ -66,7 +66,7 @@ - (void)testNSSecureCodingOfMMEEvent {
      MMEEvent *unarchived = [unarchiver decodeObjectOfClass:MMEEvent.class forKey:NSKeyedArchiveRootObjectKey];
 
      XCTAssertNotNil(unarchived);
-     XCTAssert(unarchived == event);
+     XCTAssertEqualObjects(unarchived, event);
```
dmiluski, line 69: Test Fix] testNSSecureCodingOfMMEEvent

Problem:
Comparison of objects fails

Why:
original vs unarchived failed b/c we are comparing the address of two objects which will be different. For object comparison, Asserts must use EqualObjects for expected comparison.

Change:
Make comparison with XCTAssertEqualObjects

Tests/MMEEventsManagerTests.m
```diff
@@ -164,8 +164,6 @@ - (void)testTimerReachedWithEventQueued {
     
     [self.eventsManager pushEvent:locationEvent];
     XCTAssert(self.eventsManager.eventQueue.count > 0);
-    
-    XCTAssert(self.eventsManager.eventQueue.count == 0);
```
dmiluski, line 168: Problem:
Event manager queue remains at zero

Regression:
https://github.com/mapbox/mapbox-events-ios/commit/876b4f2f5642ca28d5a066deaab775077d0cccf4#diff-b19b2c9fb2b4afec92d03b42c184b518L180

Why:
With the removal of timerManager, the events manager is no longer able to be manually flushed with trigger.

Change:
- Remove last comparison as it is no longer needed.

Tests/MMELocationManagerTests.m
```diff
@@ -18,14 +18,46 @@ @interface MMELocationManager (Spec) <CLLocationManagerDelegate>
 
 @end
 
+// MARK: - Mock
+
+@interface MockCLLocationManager: CLLocationManager
```
dmiluski: TODO: Remove. This was an attempt at mocking out location manager method calls which ran into the issue of not working due to a cached location in the location manager leaving distance filtering flaky.

Tests/MMELocationManagerTests.m
```diff
@@ -42,31 +74,6 @@ - (void)testLocationManagerStartsMonitoringRegions {
     XCTAssert(self.locationManager.monitoredRegions > 0);
 }
 
-- (void)testRegionDistanceFilter {
```
dmiluski, line 45: Invalid Test] testRegionDistanceFilter

Problem:
testRegionDistanceFilter is looking to validate that in the instance of lots of micro location updates, that we’re not overly updating our regions.

Problem:
This test is making comparisons of state synchronously, while the action is asynchronous meaning the comparisons are invalid.

Why:
https://github.com/mapbox/mapbox-events-ios/blob/master/MapboxMobileEvents/MMELocationManager.m#L215

Docs:
 *      This is done asynchronously and may not be immediately reflected in monitoredRegions.
 */
- (void)startMonitoringForRegion:(CLRegion *)region API_AVAILABLE(ios(5.0), macos(10.8)) API_UNAVAILABLE(watchos, tvos);

Approach:
Attempted to inject a spy LocationManager to detect when this was updated, unfortunately this was indeterminant due to LocationManager caching last location. Making this dependent on any other location manager which may have run prior.

Change:
- Remove Test

Alternate Options:
Fully extract out Locationmanager behind a protocol and pass that interface. This would require some implementation behaviors which may not fully match CLLocationManager.

Tests/MMEMetricsManagerTests.m
```diff
@@ -60,11 +60,11 @@ - (void)testAttributesExcludesNullIsland {
 }
 
 - (void)testAttributesDoesntExcludesNullFloat {
-    CLLocation *notNullIsland = [[CLLocation alloc] initWithLatitude:0.4 longitude:0.2];
+    CLLocation *notNullIsland = [[CLLocation alloc] initWithLatitude:4 longitude:2];
```
dmiluski, line 63: Problem:
updateCoordinate filters out coordinates that are `close` to null island.

Because the test coordinates are < 0.5, they are rounded to (0.0, 0.0) leaving the test assuming it wasn’t updated.

Change:
Bump to coordinates (4,2) which will not be rounded down to 0,0

Tests/Categories/NSBundle+TestBundle.h
```diff
@@ -0,0 +1,12 @@
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface NSBundle (TestBundle)
```
dmiluski, line 5: Goal:
No Need for Host App bundling. We are able to access test bundle with helper category.

Change:
+ Add NSBundle category exposing TestBundle reference
+ Include Test files in this bundle which are accessible regardless of app target allowing for fewer target memberships

Tests/Utilities/MMEServiceFixture.m
```diff
@@ -90,8 +91,6 @@ - (BOOL)waitForConnectionWithTimeout:(NSTimeInterval)timeout error:(NSError **)e
         goto exit;
     }
 
-    NSLog(@"%@ waiting for connection with timeout: %1.2fs", NSStringFromClass(self.class), timeout);
```
dmiluski: TODO: Return this log

MapboxMobileEvents/Categories/NSUserDefaults+MMEConfiguration_Private.h
```diff
@@ -99,7 +99,7 @@ static NSTimeInterval const MMEStartupDelayMaximum = 100;
 // MARK: -
 
 @interface NSUserDefaults (MMEConfiguration_Private)
-@property(nonatomic,readonly) NSString *mme_clietnId;
+@property(nonatomic,readonly) NSString *mme_clientId;
```
alfwatt, line 102: @dmiluski I make an embarrassing number of these errors, ask my proofreader @rclee

Tests/MMEEventTests.m
alfwatt: better version of my fix, picking this one

Tests/MMEEventsManagerTests.m
alfwatt: This might be a test regression, without that TimerManager fake it may be duplicative of another test.

alfwatt (commented):
Pending comments

dmiluski:
Closing as relevant commits were already cherry picked to:
https://github.com/mapbox/mapbox-events-ios/pull/202

PR #211, opened 2020-06-08
Extension Example: Today Widget

dmiluski:
## 🥅  Goal:
Show that MapboxMobileEvents SDK is able to be built/run in an extension.

## 🗺️  Steps:
1. Create Extension which depended on Events SDK
2. Include required permission keys in info.plist
3. Start SDK with mock values
4. Agree to permissions, and also set simulator to use bicycle ride simulation.

## 📱 Screenshot
Demo Today Extension including MKMapView as demonstration for the possibility of including a map/telemetry gathering in extension

| Extension Demo GUI |
| --- |
| Screen Shot 2020-06-08 at 4 27 07 PM |

## 📓  Notes:

1. Although there is a warning, it is not a deal breaker when working in extensions:

There is nothing preventing the SDK/Extension from building/running.

> ld: warning: linking against a dylib which is not safe for use in application extensions: /Users/danemiluski/Library/Developer/Xcode/DerivedData/MapboxMobileEvents-gtwngrjytwknvkfqciqcsoxcqjvr/Build/Products/Debug-iphonesimulator/MapboxMobileEvents.framework/MapboxMobileEvents

To enable friendlier App Extension usage though, we can signify we support app extension API only via this setting, which requires we don't require accessing SharedApplication api.

| Project App Extension Friendly Option | Errors Shown if chosen |
| --- | --- |
| Screen Shot 2020-06-08 at 5 01 26 PM | Screen Shot 2020-06-08 at 5 11 37 PM |

This can be worked around by injecting a protocol conforming wrapped item for UIApplication or NSExtension handling similar to what was done in the [beta+refactor branch](https://github.com/mapbox/mapbox-events-ios/blob/dane/beta%2Brefactor/MapboxMobileEvents/MMEUIApplicationWrapper.m#L4) did to enable testing, of which extensions are unable to support the longer running tasks. If deemed useful though, extensions could still send backtround URLSession Tasks via a background URLSession config.

2. Upon Testing, I did notice that the NSUserDefault persistent keys did not transfer between extension/app.  Will follow up to see if an app group in the TestHost enables the sharing.

### DIFF
--- a/MapboxMobileEvents.xcodeproj/project.pbxproj
+++ b/MapboxMobileEvents.xcodeproj/project.pbxproj
@@ -203,6 +203,10 @@
 		CF93A6582241EDD500BBA199 /* MMECertPin.m in Sources */ = {isa = PBXBuildFile; fileRef = CF93A6562241EDD500BBA199 /* MMECertPin.m */; };
 		CF93A6592241EDD500BBA199 /* MMECertPin.m in Sources */ = {isa = PBXBuildFile; fileRef = CF93A6562241EDD500BBA199 /* MMECertPin.m */; platformFilter = ios; };
 		CF93A65B2241EDD500BBA199 /* MMECertPin.h in Headers */ = {isa = PBXBuildFile; fileRef = CF93A6572241EDD500BBA199 /* MMECertPin.h */; };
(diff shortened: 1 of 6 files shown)


dmiluski:
Closing to verify on top of previously merged extension work:
https://github.com/mapbox/mapbox-events-ios/pull/205

As well as to clean up extension creation

PR #213, opened 2020-06-16
Cherrypick - Reformed UA Fix

rclee:
Cherry pick to include in `0.10.3`

> Check for bundle Ids as well, better filtering and sorting, for more completely consistently reformed user agent strings.

ref: https://github.com/mapbox/mobile-telemetry/issues/529


dmiluski (approved):

misterkerry (approved):
short but sweet....lgtm

MapboxMobileEvents/Categories/NSUserDefaults+MMEConfiguration.m
```diff
@@ -291,19 +291,27 @@ - (NSString *)mme_userAgentString {
...
+                [loadedMapboxBundleIds addObject:loaded.bundleIdentifier];
             }
         }
+        
+        // sort the bundleIdentifiers, then use them to build the User-Agent string
+        NSArray *sortedBundleIds = [loadedMapboxBundleIds sortedArrayUsingDescriptors:@[[NSSortDescriptor sortDescriptorWithKey:nil ascending:YES]]];
+        for (NSString *bundleId in sortedBundleIds) {
+            NSBundle *loaded = [NSBundle bundleWithIdentifier:bundleId];
+            NSString *uaFragment = [NSString stringWithFormat:@" %@/%@ (%@; v%@)",
+                loaded.infoDictionary[(id)kCFBundleNameKey],
```
alfwatt: `loaded.infoDictionary[(id)kCFBundleNameKey] ?: loaded.bundlePath.lastPathComponent. stringByDeletingPathExtension` ← `kCFBUndleNameKey` not always set

alfwatt (changes requested):
Small change to deal with a known null issue

MapboxMobileEvents/Categories/NSUserDefaults+MMEConfiguration.m
```diff
@@ -290,20 +290,30 @@ - (NSString *)mme_userAgentString {
...
+                [loadedMapboxBundleIds addObject:loaded.bundleIdentifier];
             }
         }
+        
+        // sort the bundleIdentifiers, then use them to build the User-Agent string
+        NSArray *sortedBundleIds = [loadedMapboxBundleIds sortedArrayUsingDescriptors:@[[NSSortDescriptor sortDescriptorWithKey:nil ascending:YES]]];
+        for (NSString *bundleId in sortedBundleIds) {
+            NSBundle *loaded = [NSBundle bundleWithIdentifier:bundleId];
+            NSString *uaFragment = [NSString stringWithFormat:@" %@/%@ (%@; v%@)",
+                loaded.infoDictionary[(id)kCFBundleNameKey] ?: loaded.bundlePath.lastPathComponent.stringByDeletingPathExtension,
```
dmiluski, line 309: Why delete the extension?

MapboxMobileEvents/Categories/NSUserDefaults+MMEConfiguration.m
```diff
@@ -290,20 +290,30 @@ - (NSString *)mme_userAgentString {
             NSBundle.mme_mainBundle.mme_bundleVersionString,
             NSBundle.mme_mainBundle.bundleIdentifier,
             NSBundle.mme_mainBundle.infoDictionary[(id)kCFBundleVersionKey]];
```
dmiluski, line 292: Non-blocker for the time being.

In case you want to limit code duplication and the potential for missing a fallback value, you might implement this as a NSBundle category.

Note:
```
@interface NSBundle (MapboxMobileEvents)

/*! Bundle Short version (eg. "1.0.2")
@property (nonatomic, readonly) NSString *shortVersion;

/*! The version number of the bundle. (build number eg. "1439") */
@property (nonatomic, readonly) NSString *version;

/*! The human-readable name of the bundle. */
@property (nonatomic, readonly) NSString *name;

@end

@implementation NSBundle (MapboxMobileEvents)

// implementation here, providing defaults where necessary so others don't have to

@end
```

Then a matching implementation which provides these default values so we don't need to duplicate between these top/bottom sections.

This approach would then be able to provide testable expectations by leveraging a mock bundle which provides values you might check for problematic values such as nil, or special characters.

MapboxMobileEvents/Categories/NSUserDefaults+MMEConfiguration.m
rclee: AFAIK this is to prevent a `null` value being passed through. @alfwatt for more context

MapboxMobileEvents/Categories/NSUserDefaults+MMEConfiguration.m
alfwatt: If we were using them in more than one place I might consider adding the abstraction, since this is the only spot we look at these properties I'd rather have all the logic in-line in a single function.

Because Obj-C has no namespaces, we prefix all our Category methods with `mme_` to be safe and not contribute to the `first` problem of dynamically extending class namespaces

MapboxMobileEvents/Categories/NSUserDefaults+MMEConfiguration.m
alfwatt: /long/path/to/MapboxMobileEvents.framework → MapboxMobileEvents

alfwatt: Which is typically what kCFBundleNameKey would be in English, the key exists so it can be localized

alfwatt (approved):
Thanks!

PR #217, opened 2020-06-17
Exposes Private Headers via a Development Target to both Swift and Obj-C test apps

alfwatt:
- Enables deprecation warnings on the test target


rclee (approved):

misterkerry (approved):
compiling nicely.

PR #219, opened 2020-06-17
Add UA filter

rclee:
Simple change to remove special characters from our user agents. `(),/:;<=>?@[]{}"\` are being removed from every string used to build a user agent.

### DIFF
--- a/MapboxMobileEvents/Categories/NSString+MMEVersions.m
+++ b/MapboxMobileEvents/Categories/NSString+MMEVersions.m
@@ -46,6 +46,11 @@ typedef enum {
 
 @implementation NSString (MMEVersions)
 
+- (NSString *)mme_stringByRemovingNonUserAgentTokenCharacters {
+return [[self componentsSeparatedByCharactersInSet:[NSCharacterSet characterSetWithCharactersInString:@" (),/:;<=>?@[]{}\"\\"]]
+                       componentsJoinedByString:@""];
+}
+
 + (NSRegularExpression *)mme_semverExpression {
     static NSRegularExpression *semverExpression = nil;
     static dispatch_once_t onceToken;
--- a/MapboxMobileEvents/Categories/NSString+MMEVersions.m
+++ b/MapboxMobileEvents/Categories/NSString+MMEVersions.m
@@ -47,7 +47,7 @@ typedef enum {
 @implementation NSString (MMEVersions)
 
 - (NSString *)mme_stringByRemovingNonUserAgentTokenCharacters {
-return [[self componentsSeparatedByCharactersInSet:[NSCharacterSet characterSetWithCharactersInString:@" (),/:;<=>?@[]{}\"\\"]]
+return [[self componentsSeparatedByCharactersInSet:[NSCharacterSet characterSetWithCharactersInString:@"(),/:;<=>?@[]{}\"\\"]]
                        componentsJoinedByString:@""];
 }
 
--- a/MapboxMobileEvents/Categories/NSUserDefaults+MMEConfiguration.m
+++ b/MapboxMobileEvents/Categories/NSUserDefaults+MMEConfiguration.m
@@ -286,10 +286,10 @@ NS_ASSUME_NONNULL_BEGIN
     dispatch_once(&onceToken, ^{
 
         userAgent = [NSString stringWithFormat:@"%@/%@ (%@; v%@)",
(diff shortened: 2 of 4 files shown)


MapboxMobileEvents/Categories/NSString+MMEVersions.m
```diff
@@ -46,6 +46,11 @@ - (NSString *)mme_semverBuildStringComponent;
 
 @implementation NSString (MMEVersions)
 
+- (NSString *)mme_stringByRemovingNonUserAgentTokenCharacters {
+return [[self componentsSeparatedByCharactersInSet:[NSCharacterSet characterSetWithCharactersInString:@" (),/:;<=>?@[]{}\"\\"]]
```
dmiluski: Non-Blocker/Question:
Is there any negative side effects in removing these values? Would it cause any problems when we went to look up a given value?

If there isn't. **Ignore**. 

But if there is:
Rather than remove them, what if we encoded them like a url string? This way, we could still get any characters by reverse mapping it?

Where a space " " was mapped to a url encoded " %20"

MapboxMobileEvents/Categories/NSUserDefaults+MMEConfiguration.m
```diff
@@ -286,10 +286,10 @@ - (NSString *)mme_userAgentString {
     dispatch_once(&onceToken, ^{
 
         userAgent = [NSString stringWithFormat:@"%@/%@ (%@; v%@)",
-            NSBundle.mme_mainBundle.infoDictionary[(id)kCFBundleNameKey] ?: NSBundle.mme_mainBundle.bundlePath.lastPathComponent.stringByDeletingPathExtension,
```
dmiluski, line 291: Non-Blocker:
This may be of little value given the life expectancy of this repo.

There's a a number of repetitive calls here where a new piece of information could forget to include the additional stripping. To prevent this, the stripping could take place on the collection rather than individually?

Option: Strip After setting up array
```
// Collection of Input Values
NSArray * values = [
   [NSBundle.mme_mainBundle objectForInfoDictionaryKey:(id)kCFBundleNameKey] ?: NSBundle.mme_mainBundle.bundlePath.lastPathComponent.stringByDeletingPathExtension,
   NSBundle.mme_mainBundle.mme_bundleVersionString,
    NSBundle.mme_mainBundle.bundleIdentifier,
    NSBundle.mme_mainBundle.infoDictionary[(id)kCFBundleVersionKey]
];

// Strip Out invalid characters to form valid arguments
NSMutableArray *arguments = [NSMutableArray arrayWithCapacity:[values count]];
[self enumerateObjectsUsingBlock:^(id obj, NSUInteger idx, BOOL *stop) {
        [result addObject:[(NSString*)obj mme_stringByRemovingNonUserAgentTokenCharacters]];
}];
userAgent = [NSString stringWithFormat:@"%@/%@ (%@; v%@)", arguments];
```

Feel free to ignore though, this may see very little value given the differences between the two sections here.

MapboxMobileEvents/Categories/NSString+MMEVersions.m
alfwatt: I think spaces can be maintained "My Great Application" isn't an issue but "My App Build (123/123/123/;;;/.)" is

alfwatt (changes requested):
We can tolerate spaces in those fields, just not the delimiter tokens
Would be great to have a test with that one User-Agent string that started this, maybe with an example app name substituted

MapboxMobileEvents/Categories/NSString+MMEVersions.m
alfwatt: There *are* encoding rules for the UA string if you look at the HTTP RFC, they can be escaped with `\` to retain them, but parsing that is problematic so it's easier for us to just compress out any invalid tokens that we've given

dmiluski: Makes sense. Thanks for the clarification

MapboxMobileEvents/Categories/NSUserDefaults+MMEConfiguration.m
dmiluski: Removing comment. this is clutter.

dmiluski (approved):

alfwatt:
Looks good, if that merge commit does OK in CI I think we're good to merge these into my branch and review that.

alfwatt (approved):

PR #220, opened 2020-06-18
🔪

alfwatt:
'Fetch me a sword.' And they brought a sword before the king.

--

This change re-structures the Xcode Projects to make builds with Carthage simpler and more reliable and to provide a `(Development)` target which includes the `PrivateHeaders` from the project so that the tests can be relocated to a different project and example apps, stagecoach, etc. can use these APIs


alfwatt:
This will require some workflow changes:
- For development you need to open `MapboxMobileEventsTest.xcodeproj`
- Tests are run as part of the `MMETestHost` scheme not the framework schemes
  - Bitrise will need to be reconfigured for that
- `MapboxMobileEvents (Development)` has all the project headers in the `PrivateHeaders` so that they can be accessed by the tests, it needs to stay synched with the release target

dmiluski:
Could we include the goal/rationale for why we are splitting this project up?

I believe it's related to building the dependency via carthage, but I wanted to double check so I had the correct information to review.

One thing I wonder is:
1. Given this change, will we also be able to evaluate code coverage given the tests are split into a separate project?
2. What is the expected Bitrise change to support?

I fear that we are making a big project file change last minute to accommodate which could add risk to the overall release.

One change I found that supported building successfully in carthage was to remove the (Release) scheme.
https://github.com/mapbox/mapbox-events-ios/compare/hack/releaseBuildToyingAround?expand=1

And tested with Cartfile
```
binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" ~> 5.8
github "mapbox/mapbox-events-ios" "hack/releaseBuildToyingAround"
```

And carthage running
```
danemiluski@C02ZC5RNLVDL mobile-telemetry-example-ios % carthage update --platform ios

*** Downloading binary-only framework Mapbox-iOS-SDK at "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json"
*** Fetching mapbox-events-ios
*** Checking out mapbox-events-ios at "6f587bdb0668bb261b93df1f210c0ee18708e638"
*** xcodebuild output can be found in /var/folders/ks/ffsh1zgn35s56mynz581zzt40000gq/T/carthage-xcodebuild.5dwXdH.log
*** Downloading binary-only framework Mapbox-iOS-SDK at "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json"
*** Building scheme "MapboxMobileEvents" in MapboxMobileEvents.xcodeproj
```

Although I understand we're trying to provide both abilities to provide Release/Development options, I wonder if perhaps we could make those changes separate from the release to unblock it's building/testing?

Additionally, I agree that the widget can be removed, or at minimum moved to an example directory separate from the main project to keep the project clean and focused on its responsibilities. I also see that TestHost can be removed as a requirement to build/run tests. Although both of these changes would take additional changes which could put the release at risk timewise.

misterkerry:
Maybe we can consider moving any extension to a separate Xcode project within the workspace to provide and an example that is decoupled from MME.

rclee (approved):

dmiluski (approved):

PR #235, opened 2020-09-03
Add CocoaPods and Carthage integration tests

frederoni:
Added tests for integrating this library with CocoaPods and Carthage.

What the tests verify is that the podspec is valid and we are able to integrate it into a project and that there's a valid shared scheme Carthage can use to integrate the framework and we can compile an application using one of package managers.

cc @alfwatt @mr1sunshine @nagineni

### DIFF
--- a/Tests/Integration/Carthage/project.yml
+++ b/Tests/Integration/Carthage/project.yml
@@ -1,0 +1,13 @@
+name: CarthageTest
+options:
+  bundleIdPrefix: com.mapbox.common.events.carthage
+targets:
+  CarthageTest:
+    type: application
+    platform: iOS
+    deploymentTarget: "9.0"
+    sources: [Sources]
+    dependencies:
+      - framework: Carthage/Build/iOS/MapboxMobileEvents.framework
+    settings:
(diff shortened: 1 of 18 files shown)


Tests/Integration/Carthage/project.yml
```diff
@@ -0,0 +1,13 @@
...
+targets:
+  CarthageTest:
+    type: application
+    platform: iOS
+    deploymentTarget: "9.0"
+    sources: [Sources]
+    dependencies:
+      - framework: Carthage/Build/iOS/MapboxMobileEvents.framework
+    settings:
+        DEVELOPMENT_TEAM: "GJZR2MEM28"
```
mr1sunshine, line 13: Is that `DEVELOPMENT_TEAM` correct?

mr1sunshine (approved):
LGTM

Tests/Integration/Carthage/project.yml
frederoni: It's not being used for building and testing on a simulator, but it is the correct team id.

PR #237, opened 2020-09-04
[src] Add accuracy authorization field to appUserTurnstile event

mr1sunshine:


MapboxMobileEvents/Categories/CLLocationManager+MMEMobileEvents.m
```diff
@@ -1,14 +1,25 @@
 #import "CLLocationManager+MMEMobileEvents.h"
-#import "MMEConstants.h"
+#import "../MMEConstants.h"
```
frederoni: All of these headers are flattened and would end up in the same directory so dropping `../` must work.

alfwatt (approved):
Looks good

PR #238, opened 2020-09-05
[ci] Add CircleCI build job for XCode 12 (beta 6)

mr1sunshine:
In order to build SDK and run test against iOS 14 and at the same
time keep ability to verify that everything works with XCode 11
we are adding additional job for XCode 12

There is an issue with Carthage and XCode 12 (beta 6) which requires
us to use custom xcconfig file to exclude arm architectures.


Tests/MMEAPIConfigTests.m
```diff
@@ -49,6 +49,8 @@ - (void) test001_TestHostConfigURL {
 }
 
 - (void) test002_StaringConfigUdpate {
+    XCTSkipIf(([[NSProcessInfo processInfo] isOperatingSystemAtLeastVersion:(NSOperatingSystemVersion){.majorVersion = 14, .minorVersion = 0, .patchVersion = 0}]), @"Skip, since it is not working on iOS14");
```
frederoni: As a reminder to ourselves, can you add in the description what `it` means? `CFSocketInvalidate` crashes with `EXC_GUARD` on iOS 14

frederoni (approved):
LGTM % nit

Tests/MMEAPIConfigTests.m
mr1sunshine: > As a reminder to ourselves, can you add in the description what `it` means? `CFSocketInvalidate` crashes with `EXC_GUARD` on iOS 14

Good point.

PR #239, opened 2020-09-08
Offload heavy request builder work

frederoni:
Dispatch to a global queue with default priority when calling `-[MMEAPIClient requestForConfiguration]`

Using 12 assorted dependencies
<details><summary>Dependencies</summary>

```bash
MapboxMobileEvents
MapboxAccounts
MapboxCommon
MapboxDirections
MapboxGeocoder.swift
MapboxSpeech
MapboxStatic.swift
Fingertips
Turf
Alamofire
SwiftyJSON
FacebookSDK
```

</details>

indicated that initializing the MMEEventManager and MMEAPIClient spent ~100ms on the main thread and with this change, we spend ~3ms on the main thread.

Before

After

cc @alfwatt @nagineni @mr1sunshine

### DIFF
--- a/MapboxMobileEvents/MMEAPIClient.m
+++ b/MapboxMobileEvents/MMEAPIClient.m
@@ -138,6 +138,11 @@ int const kMMEMaxRequestCount = 1000;
 }
 
 // MARK: - Configuration Service
+void MMEDispatchMainIfNeeded(void (^block)(void))
+{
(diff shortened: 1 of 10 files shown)


alfwatt (commented):
Good fix for the specific issue with initial config loading and setup.

We should fix this more generally for all our networking and disk IO tasks at some point.

alfwatt (approved):

PR #243, opened 2020-09-09
Replace @module import with #import for Obj-C++ client support

datwelk:
When using MapboxEvents in combination with an Objective-C++ class, you will get the following error:

```
Use of @import when modules are disabled
```

This is due to the `@import` statement inside the framework's exposed `NSUserDefaults+MMEConfiguration.h` header. Changing it to the traditional angle bracket import fixes the issue.


frederoni:
@datwelk thank you for your contribution. Would you mind doing a force push to start the CI bots?

datwelk:
@frederoni Happy to contribute! Amended my last commit and pushed, CircleCI is now pending.

frederoni (approved):

PR #248, opened 2020-09-10
[location] Add location counters

mr1sunshine:
As part of telemetryMetrics event we would like to collect as well
    the following counters:

    locationsFG: count of locations received with the app in the foreground
    locationsBG: count of locations received with the app in the background
    locationsAX: count of locations received with approximate values
    locationsHF: count of locations dropped due to horizontal accuracy filter
    locationsTX: count of locations converted into events and enque


alfwatt (changes requested):
- [x] Needs tests for setting values, saving and restoring the metrics object
- [x] Implement counter for the events dropped by the HAO filter

Tests/MMEMetricsManagerTests.mm
```diff
@@ -270,6 +270,36 @@ + (NSString *)pendingMetricsEventPath;
                 [manager loadPendingTelemetryMetricsEvent] should be_nil;
             });
         });
+        context(@"when incrementing location events related counters", ^{
```
alfwatt: Can we move these to XCtests? It's OK if there are some tests in cedar still but I don't want to add any more cedar tests to the project.

We also need tests to cover the persistence of these values, this is the only event we write to storage and restore so we should double check

alfwatt (changes requested):

alfwatt (approved):
Thanks @mr1sunshine !

PR #249, opened 2020-09-11
[events] Send digest header value to the location event

nagineni:
Send digest header value to location events in order to relate events which were generated with various configurations.

### DIFF
--- a/MapboxMobileEvents/MMEAPIClient.m
+++ b/MapboxMobileEvents/MMEAPIClient.m
@@ -175,6 +175,29 @@ void MMEDispatchMainIfNeeded(void (^block)(void))
     }
 }
 
++ (nullable NSString *)parseDigestHeader:(NSString *)digestHeader {
+    NSString *digest = nil;
+    if (digestHeader) {
+        // Look for 'SHA-256' hash in case of multiple different digest values
+        NSArray *kvs = [digestHeader componentsSeparatedByString:@","];
+        if (kvs.count >= 2) {
+            for (NSString *field in kvs) {
+                NSArray *keyValue = [field componentsSeparatedByString:@"SHA-256="];
+                if (keyValue.count == 2) {
+                    digest = keyValue[1];
+                }
+            }
+        } else {
+            // In case of single digest value, look for the 'SHA-256' hash, and
+            // fallback to the value received from the config service assuming
+            // that the Digest: header comes from a 1.0 config service has just
+            // a hash value.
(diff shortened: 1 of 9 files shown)


MapboxMobileEvents/MMEAPIClient.m
```diff
@@ -207,6 +207,29 @@ - (void)processRequest:(NSURLRequest *)request {
             }
 
             NSUserDefaults.mme_configuration.mme_configUpdateDate = MMEDate.date;
+
+            NSString *digest = nil;
+            NSString *digestHeader = httpResponse.allHeaderFields[@"Digest"];
```
mr1sunshine: nit: Should we use here `valueForHTTPHeaderField` method?

mr1sunshine (approved):
LGTM with small nit.

MapboxMobileEvents/MMEAPIClient.m
```diff
@@ -207,6 +207,29 @@ - (void)processRequest:(NSURLRequest *)request {
...
+            if (digestHeader) {
+                // Look for 'SHA-256' hash in case of multiple different digest values
+                NSArray *digestValues = [digestHeader componentsSeparatedByString:@","];
+                if (digestValues.count >= 2) {
```
frederoni: NSArray is now `key=value`, right?
could we do something like (with some additional checks):
```
    NSArray *kvs = [digest componentsSeparatedByString:@","];
    NSMutableDictionary *dict = [NSMutableDictionary dictionary];
    for (NSString *field in kvs) {
        NSArray *keyValue = [field componentsSeparatedByString:@"="];
        dict[keyValue[0]] = keyValue[1];
    }
```
to get a cleaner structure like:
```
{
    Digest = abcd;
    "SHA-256" = 1234;
}
```
or are there too many variations in the data?

MapboxMobileEvents/MMEAPIClient.m
nagineni: 'valueForHTTPHeaderField:' is only available on iOS 13.0 or newer, so I am not sure whether we should use that or not.

MapboxMobileEvents/MMEAPIClient.m
nagineni: Unfortunately, the current config service is not sending the `digest` header as `<key>=<value>` format, and that's the reason we need handle the different corner cases to fallback to the value(just hash) received from the config service.

PR #250, opened 2020-09-11
Refactor project and add support for SPM

frederoni:
Refactored the project to address the following issues

- Added support for Swift Package Manager
- Restructured folders to simplify SPM setup
- Folded MapboxMobileEventsTests.xcodeproj into MapboxMobileEvents.xcodeproj for a more convenient development workflow using only one project file
- Removed all shared schemes except `MapboxMobileEvents` to avoid confusion and speed up Carthage builds by a factor of 3x
- Removed the static libMapboxMobileEvents scheme since it's no longer being used by gl-native
- Removed the name header script since we are no longer dual-linking
- Added codecov to CircleCI setup so Bitrise is no longer required

The SPM setup  is running all XCTests
MapboxMobileEvents.xcodeproj is running all Cedar _and_ XCTests

cc @alfwatt @nagineni @mr1sunshine


.circleci/config.yml
```diff
@@ -42,3 +46,10 @@ jobs:
           command: |
             if [ "<< parameters.xcode >>" = "12.0.0" ]; then export XCODE_XCCONFIG_FILE=~/project/xcode-12-fix.xcconfig; fi
             Tests/Integration/test_carthage.sh
+      - when:
+          condition:
+            equal: [ "11.7.0", << parameters.xcode >> ]
```
mr1sunshine, line 51: Is there any problem with XCode 12 code coverage?

MapboxMobileEvents.podspec
```diff
@@ -21,7 +21,7 @@ Pod::Spec.new do |s|
 
   # ――― Platform Specifics ――――――――――――――――――――――――――――――――――――――――――――――――――――――― #
 
-  s.ios.deployment_target = "8.0"
+  s.ios.deployment_target = "9.0"
```
mr1sunshine, line 24: I wonder if we should have bump of minimum deployment target in the separate PR/commit?

Package.swift
```diff
@@ -0,0 +1,38 @@
+// swift-tools-version:5.2
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+
+let package = Package(
+    name: "MapboxMobileEvents",
+    platforms: [
+        .iOS(.v10)
```
mr1sunshine: Does it matter which version of `iOS` we set here?

.circleci/config.yml
frederoni: Not that I know of but since Xcode 12 is still in beta and we run the same tests as with Xcode 11, I figured merging multiple reports would be excessive at this point.

Tests/MapboxMobileEventsTests/MMEAPIConfigTests.m
```diff
@@ -95,6 +95,7 @@ - (void) test004_ShortUpdateInterval {
     XCTAssertNil(configError);
 }
 
+#if !SWIFT_PACKAGE
```
mr1sunshine, line 99: What is the problem with that test? Worth adding a comment?

Tests/MapboxMobileEventsTests/MMECertPinTests.m
```diff
@@ -83,7 +83,8 @@ -(void)test002_checkCOMHashCount {
     NSArray *comHashes = NSUserDefaults.mme_configuration.mme_certificatePinningConfig[@"events.mapbox.com"];
     XCTAssert(comHashes.count == 4);
 }
-    
+
+#if !SWIFT_PACKAGE
```
mr1sunshine, line 87: the same: should we add a comment on why we are commenting out?

Tests/MapboxMobileEventsTests/MMEEventsManagerTests.m
```diff
@@ -75,6 +75,7 @@ - (void)testPausesWithWhenInUseAuthAndBackgrounded {
     XCTAssert(self.eventsManager.paused == YES);
 }
 
+#if !SWIFT_PACKAGE
```
mr1sunshine, line 79: the same here?

.circleci/config.yml
```diff
@@ -23,17 +23,21 @@ jobs:
...
       - checkout
-      - run: make build
-      - store_artifacts:
-          path: build/mapbox-events-ios-static.zip
+      - run:
+          name: Test SPM
+          command: |
+            # xcodebuild will try to use the Package manifest if it cannot find a project 
+            mkdir temp && mv MapboxMobileEvents.xcodeproj temp
+            xcodebuild -scheme MapboxMobileEvents test -destination "platform=iOS Simulator,name=iPhone 11,OS=latest"
```
nagineni, line 31: Does it make sense to specify the platform value through the build parameter?

mr1sunshine (approved):
LGTM with a couple of comments about why we are disabling tests.

nagineni (approved):

.circleci/config.yml
frederoni: Yes, as a follow-up because then it should also propagate to CocoaPods and Carthage tests.

Package.swift
frederoni: Yes, good catch. Lowered to v9.

Tests/MapboxMobileEventsTests/MMEAPIConfigTests.m
frederoni: Support for resources was added in Swift 5.3 but we currently only require 5.2. 5.2 is included in Xcode 11 and 5.3 in 12.

PR #254, opened 2020-09-16
[src] Use lazy initializers for MMELocationManager and CLLocationManager

mr1sunshine:
Starting from iOS14 in order to get locationAuthorization(and
accuracyAutorization) we need to call instance methods of
CLLocationManager. Since it is possible for SDKs to send userTurnstile
event before finishing construction of MMEEventManager (we have
`initializtion` callback in `initializeWithAccessToken`) we might end up
in situation when CCLocationManager is not created yet.

To avoid behaviour above we introduce two lazy initializers for:
- MMELocationManager in MMEEventManager
- CCLocationManager in MMELocationManager
so both location managers will be always available whenver we try to
access to them.

After this change there is no need to set CCLocationManager pointer to
nil when we call `stopUpdatingLocation` since location manager now is
always available.

### DIFF
--- a/Sources/MapboxMobileEvents/MMEEventsManager.m
+++ b/Sources/MapboxMobileEvents/MMEEventsManager.m
@@ -126,7 +126,6 @@ NS_ASSUME_NONNULL_BEGIN
                 }
 
                 strongSelf.paused = YES;
-                strongSelf.locationManager = [[MMELocationManager alloc] init];
                 strongSelf.locationManager.delegate = strongSelf;
                 [strongSelf resumeMetricsCollection];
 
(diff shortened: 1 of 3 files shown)


Sources/MapboxMobileEvents/MMEEventsManager.m
```diff
@@ -126,8 +126,7 @@ - (void)initializeWithAccessToken:(NSString *)accessToken userAgentBase:(NSStrin
                 }
 
                 strongSelf.paused = YES;
-                strongSelf.locationManager = [[MMELocationManager alloc] init];
-                strongSelf.locationManager.delegate = strongSelf;
+                strongSelf.locationManager.delegate = self;
```
frederoni: s/self/strongSelf to avoid retain cycle

frederoni (approved):
LGTM % the comment but let's wait for @alfwatt to 👀

Sources/MapboxMobileEvents/MMEEventsManager.m
```diff
@@ -572,6 +571,15 @@ - (void)pushEvent:(MMEEvent *)event {
     }
 }
 
+- (id<MMELocationManager>) locationManager {
+    NSAssert(NSThread.isMainThread, @"locationManager should be accessed only from the main thread");
```
alfwatt: Doesn't this halt? I'd rather we didn't crash the hosting SDK/App in this case

alfwatt: Suggest replacing it was a `dispatch_once` block to prevent the race codition

alfwatt (approved):
Approved with suggested change to `-locationManager`

PR #257, opened 2020-09-17
Fix non-primitive assertion

frederoni:
XCTAssertNotEqual is meant to compare primitives but the objects we were comparing are not primitives.

@alfwatt @mr1sunshine


mr1sunshine (approved):

nagineni (approved):

alfwatt:
Good catch, there were a few like it. This is included in 655f50463743ad9dbe6120fedd3c4c2ddc8ae334 and 74014a6056f6f5157eeda1e128d64bb49b064d4e

PR #259, opened 2020-09-23
CFBundleShortVersionString must conform to [Major].[Minor].[Patch]

frederoni:
[CFBundleShortVersionString](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleshortversionstring) can only contain numeric characters (0-9) and periods. Each integer provides information about the build version in the format [Major].[Minor].[Patch]

cc @nagineni @alfwatt @bamx23


nagineni (approved):

alfwatt:
This isn't going to work in practice, how will we tell the difference between the version in testing? We won't know the specific version that a build is?

[`CFBUndleShortVersionString`](https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/TP40009249-111349-TPXREF113) and [`CFBundleVersion`](https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/TP40009249-102364-TPXREF106) are better described in the CoreFoundation docs

Also, what's the driver to change this? We've been using the alpha strings for a while without issues.

frederoni:
Reported from App Store Connect / iTunes Connect:
```
ERROR ITMS-90058: "This bundle is invalid. The value for key CFBundleVersion [0.11.0-beta.2] in the Info.plist file must be a period-separated list of at most three non-negative integers. Please find more information about CFBundleVersion at https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleversion"
```
So Apple’s automatic reviewing process for TestFlight submission caught this.

After the fix in this PR, @bamx23 reported that this branch is working fine.

However, I need to dig deeper and find the root cause. I have a hunch the Info.plist may be autogenerated when integrating with CocoaPods and potentially excluded when integrating with Carthage, and included as-is when integrating with SPM which scans the source folder recursively.

Need some sort of Apple Application Loader test or fastlane testflight distribution test utilizing all package managers.

PR #268, opened 2020-10-28
Only warn about semver when checking the `MGLSemanticVersionString`

alfwatt:
+ Don't include the matched alpha/beta tag in the marketing version 
+ Add the MGLSemanticVersionString to the info plist of the framework, for the user agent string
+ v0.10.6-alpha.1

Fixes: https://github.com/mapbox/mobile-telemetry/issues/703


mr1sunshine (approved):
LGTM

PR #275, opened 2021-01-20
Improve the comparison with the bundle id

tattn:
This is a change to improve the speed of user agent generation.

- Objective-C does not crash if the receiver is `nil`, and returns `false`, so nil checking can be reduced.
- Forward matching is faster for string comparisons. The bundle ID string of the mapbox SDK starts with `com.mapbox`. 

If there is a problem with the implementation, let me know. 😃


tattn:
As a result of measuring with Time Profiler, the user-agent generation is about 10% faster for our apps.

alfwatt (approved):
Looks good, the Null check might have been slightly faster but it's rare to get a null in that set

tattn:
@rclee How about this PR? If it's OK, please merge it.  😄

YoshiYazawa:
@alfwatt has this update been merged?

vayesh (approved):

vayesh:
@tattn LGTM. Could you rebase it on main and merge?

tattn:
@vesh93 Thank you for your review! It's done 👍

PR #287, opened 2021-05-21
[build] Separate job for building release artefacts

vayesh:
Release artefacts should be built once, on xcode 12. Therefore, there should be a separate job for that.


mr1sunshine (approved):
LGTM

PR #307, opened 2021-09-02
Optimize userAgentString building time

vayesh:
Optimize `mme_userAgentString` method by:
1) Using in-memory list of loaded dylibs instead of `[NSBundle allFrameworks]` method.
2) Filtering out all out-of-bundle frameworks, since it's impossible for Mapbox Frameworks to be located anywhere else.

### DIFF
--- a/Sources/MapboxMobileEvents/NSUserDefaults+MMEConfiguration.m
+++ b/Sources/MapboxMobileEvents/NSUserDefaults+MMEConfiguration.m
@@ -1,4 +1,5 @@
 #import <CommonCrypto/CommonDigest.h>
+#import <mach-o/dyld_images.h>
 
 #import "MMEConstants.h"
 #import "MMEDate.h"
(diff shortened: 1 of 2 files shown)


Sources/MapboxMobileEvents/NSUserDefaults+MMEConfiguration.m
```diff
@@ -317,12 +318,38 @@ - (NSString *)mme_userAgentString {
              [NSProcessInfo mme_processorTypeDescription]
         ];
         
+        // try to get a list of loaded dylibs
+        struct task_dyld_info tdi;
+        mach_msg_type_number_t count = TASK_DYLD_INFO_COUNT;
```
macdrevx, line 323: Why is this variable initialized to `TASK_DYLD_INFO_COUNT`. It seems like it's used to collect the count of the output from `task_info`, so could it just be initialized to 0?

macdrevx: Also, is `count` always 1 after `task_info` returns?

Sources/MapboxMobileEvents/NSUserDefaults+MMEConfiguration.m
```diff
@@ -317,12 +318,38 @@ - (NSString *)mme_userAgentString {
...
-                && loaded.bundleIdentifier != NSBundle.mme_mainBundle.bundleIdentifier) {
-                [loadedMapboxBundleIds addObject:loaded.bundleIdentifier];
+        for (uint32_t i = 0; i < infos->infoArrayCount; ++i) {
+            struct dyld_image_info *imageInfo = infoArray + i;
+
+            NSString* path = [NSString stringWithCString:imageInfo->imageFilePath encoding:NSUTF8StringEncoding];
+            // check if a framework located inside app's bundle, only those can be ours
+#if !TARGET_OS_SIMULATOR
+            NSString* mainBundlePath = NSBundle.mainBundle.bundlePath;
+            if ([path hasPrefix:mainBundlePath]) {
```
macdrevx, line 342: Why doesn't this work on simulator?

Sources/MapboxMobileEvents/NSUserDefaults+MMEConfiguration.m
vayesh: That is how the API works. Check `task_info.h` - there are many `*_INFO_COUNT` constants. To get a list of loaded dylibs, we have to pass this one.

vayesh: Also, there are some examples in `https://opensource.apple.com/source/xnu/xnu-4903.231.4/tests/task_info.c.auto.html`

macdrevx (approved):

Sources/MapboxMobileEvents/NSUserDefaults+MMEConfiguration.m
vayesh: From our experience, on simulator Frameworks are loaded not from `<app_bundle>/Frameworks`, but directly from build directories for some reason. So filtering by app_bundle path is not possible. Not sure if there is some issue with project settings or does it work as intended.

vayesh: We are more interested on optimizing the method on real devices, so decided to go `#ifdef` way here.

Sources/MapboxMobileEvents/NSUserDefaults+MMEConfiguration.m
macdrevx: Interesting. Here's a link to the source for anyone who wants to read more… https://github.com/apple/darwin-xnu/blob/2ff845c2e033bd0ff64b5b6aa6063a1f8f65aa32/osfmk/kern/task.c

Seems like it's used to indicate the size/version of the 3rd argument. There's even an interesting workaround for a specific iOS version that used this API incorrectly… https://github.com/apple/darwin-xnu/blob/2ff845c2e033bd0ff64b5b6aa6063a1f8f65aa32/osfmk/kern/task.c#L5248-L5257

PR #322, opened 2021-10-12
Fix null ptr access crash

### LINKED ISSUE #321: 🐞 [iOS] Crash: NSUserDefaults+MMEConfiguration.m - Line 338
**Configuration**
 - SDK Version 1.0.5
 - iOS Version: 14.7.1 and 14.8
 - Device: iPhone X and iPhone 11 Pro
 
**Steps to Reproduce**
We have Crashlytics reports from the App Store version of the app where it crashes in the SDK. I don't know the steps to reproduce since these are happening to our users in the wild. I can see, based on logs, that it happens at the launch of the app. 

**Logging**
Below is the Crashlytics report that we got, let me know how else I can help. 

Screen Shot 2021-10-11 at 4 33 10 PM

vayesh:
Fixes #321.

### DIFF
--- a/Sources/MapboxMobileEvents/NSUserDefaults+MMEConfiguration.m
+++ b/Sources/MapboxMobileEvents/NSUserDefaults+MMEConfiguration.m
@@ -335,6 +335,10 @@ NS_ASSUME_NONNULL_BEGIN
         for (uint32_t i = 0; i < infos->infoArrayCount; ++i) {
             struct dyld_image_info *imageInfo = infoArray + i;
 
+            if (imageInfo == NULL || imageInfo->imageFilePath == NULL) {
+                continue;
+            }
+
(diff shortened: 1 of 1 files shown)


nagineni (approved):

Sources/MapboxMobileEvents/NSUserDefaults+MMEConfiguration.m
```diff
@@ -335,6 +335,10 @@ - (NSString *)mme_userAgentString {
         for (uint32_t i = 0; i < infos->infoArrayCount; ++i) {
             struct dyld_image_info *imageInfo = infoArray + i;
 
+            if (imageInfo == 0 || imageInfo->imageFilePath == 0) {
```
mr1sunshine: Should we use `NULL` instead of `0`?

PR #350, opened 2022-09-21
Bugfix for `This method can cause UI unresponsiveness` warning

OdNairy:
Fixes iOS 16 SDK runtime warning `This method can cause UI unresponsiveness`. 
First fix part in https://github.com/mapbox/mapbox-events-ios/pull/344

Fixes: MAPSIOS-253
Fixes: CORESDK-1215

### DIFF
--- a/Sources/MapboxMobileEvents/MMELocationManager.m
+++ b/Sources/MapboxMobileEvents/MMELocationManager.m
@@ -77,9 +77,6 @@ NSString * const MMELocationManagerRegionIdentifier = @"MMELocationManagerRegion
 }
 
 - (void)startUpdatingLocation {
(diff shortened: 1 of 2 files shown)


Sources/MapboxMobileEvents/MMELocationManager.m
```diff
@@ -77,9 +77,6 @@ - (instancetype)init {
 }
 
 - (void)startUpdatingLocation {
-    if (![CLLocationManager locationServicesEnabled]) {
-        return;
-    }
```
OdNairy, line 82: We actually don't need this check because we do Apple recommended auth-based conditions later before starting locations monitoring.

evil159 (approved):

PR #349, opened 2022-09-21
Fix missing location telemetry events

OdNairy:
Due to rescheduling some telemetry event to custom dispatch queue the self.locationManager getter (that includes instantiation logic) was called on the non-main queue.
CLLocationManager requires to have NSRunLoop in the constructor thread.
Instantiation in non-main queue was an unexpected change

plus:
* minor refactoring

Fixes https://mapbox.atlassian.net/browse/CORESDK-1286

PR #352, opened 2022-11-01
Lock-free LocationManager instantiation

OdNairy:
This PR address the deadlock issue. 
Sometimes access to `MMELocationManager/locationManager` happens in the middle of the original `locationManager` getter. For example, code with access to `locationManager` can be scheduled for the next main RunLoop run, and then the getter would schedule its own internal initialization to the main queue and keep holding the `@synchronized` lock. And on the next RunLoop run when an arbitrary access code would be called that lock would stop the main queue execution without a way to break the deadlock.
Some extra context in the [internal discussion](https://mapbox.slack.com/archives/C0ACM9Q2C/p1667305455583319).

Fixes https://mapbox.atlassian.net/browse/CORESDK-1440

Related:
- https://github.com/mapbox/mapbox-events-ios/pull/349

### DIFF
--- a/Sources/MapboxMobileEvents/MMELocationManager.m
+++ b/Sources/MapboxMobileEvents/MMELocationManager.m
@@ -37,32 +37,46 @@ NSString * const MMELocationManagerRegionIdentifier = @"MMELocationManagerRegion
     _locationManager.delegate = nil;
 }
 
-- (CLLocationManager *)locationManager  {
-    @synchronized (self) {
-        if (!_locationManager) {
-            dispatch_block_t instantiateLocationManager = ^{
-                _locationManager = [[MMEDependencyManager sharedManager] locationManagerInstance];
-            };
-            if ([NSThread isMainThread]) {
-                instantiateLocationManager();
-            } else {
-                dispatch_sync(dispatch_get_main_queue(), instantiateLocationManager);
-            }
-        }
-        return _locationManager;
+- (CLLocationManager *)locationManager {
+    // Always reschedule to main queue to avoid race condition
+    if (![NSThread isMainThread]) {
+        __block CLLocationManager* manager = nil;
+        // Avoid other synhronization methods like @synchronized
+        // since the access to locationManager might happen in
+        // main thread after this function call but before
+        // mainQueue async.
+        dispatch_sync(dispatch_get_main_queue(), ^{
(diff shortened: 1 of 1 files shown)


Sources/MapboxMobileEvents/MMELocationManager.m
```diff
@@ -37,32 +37,46 @@ - (void)dealloc {
...
-        return _locationManager;
+- (CLLocationManager *)locationManager {
+    // Always reschedule to main queue to avoid race condition
+    if (![NSThread isMainThread]) {
+        __block CLLocationManager* manager = nil;
+        // Avoid other synhronization methods like @synchronized
+        // since the access to locationManager might happen in
+        // main thread after this function call but before
+        // mainQueue async.
+        dispatch_sync(dispatch_get_main_queue(), ^{
```
S2Ler, line 48: theoretically this can deadlock as well. If you sync on background thread from the main one while background thread will call this method. 

Not sure if it is possible scenario with this lib though.

S2Ler (approved):

Sources/MapboxMobileEvents/MMELocationManager.m
OdNairy: hm, you are right. Any ideas on how to fix that? 🙂

tarigo (approved):

Sources/MapboxMobileEvents/MMELocationManager.m
S2Ler: this approach seem to work, take a look
[Deadlock.zip](https://github.com/mapbox/mapbox-events-ios/files/9921720/Deadlock.zip)

S2Ler: just not sure about run loop. it is not guaranteed to be available.

OdNairy: I think it is acceptable risk. My point is that `sync` call from main thread to the background queue in not something common and should be generally avoided as a bad practice leading to the UI lags.

OdNairy: Thanks for providing a workaround. It seems to have some overhead so I will avoid it at the moment. In case of any issues reports I will reconsider my position.

PR #360, opened 2024-09-12
Remove bitcode generation

aleksproger:
We have the [following issue ](https://github.com/mapbox/mapbox-maps-ios/issues/2233)in public github, which states that clients who use Xcode 16 for AppStore uploads and use v10 Maps facing issue due to presence of bitcode in our frameworks. 

There is a workaround to delete bitcode before using our frameworks, however I think we better should remove bitcode from our v10 stack as it's deprecated already for almost 2 years. Basically, their should be no bitcode users currently as products with bitcode wasn't accepted by AppStore since Xcode 14 introduction.

So in order to unblock customers of v10 and allow to upload their apps to AppStore we must make new minor v10 releases for (mapbox-events-ios, mapbox-common, mapbox-core-maps) without bitcode

Now new binary doesn't contain bitcode
image
While the current one from `https://api.mapbox.com/downloads/v2/mapbox-events-ios/releases/ios/packages/1.0.10/MapboxMobileEvents.zip` has bitcode
image


tatiana-yan:
> new minor v10 releases

New minor v10 releases should not depend on this library, if for some reason dependency exists, it should be removed. We switched to core telemetry between v10.8 and v10.9 more than 2 years ago.
The current v10 release is v10.15

aleksproger:
> current v10 release is v10.15

Current is 10.18 and anyway while we're investigating the reasons for the existence of this dependency I suggest to merge this (without the immediate release) as it at least make this library compatible with modern hardware and CircleCI

aleksproger:
So, there is a long discussin in iOS about whether we can drop the distribution of MME, but there is a chance that somebody may use those APIs directly, to be sure let's create a release with this changes it will make the infra up-to-date, remove bitcode and add support for arm64 ios-simulator architecture

Frameworks/Cedar.xcframework/ios-arm64/Cedar.framework/Headers/CDRSpec.h
```diff
@@ -16,24 +16,30 @@ extern const __nullable CDRSpecBlock PENDING;
 #ifdef __cplusplus
 extern "C" {
 #endif
-void beforeEach(CDRSpecBlock);
+
+    void beforeEach(CDRSpecBlock);
```
tatiana-yan: nit: unexpected spaces at the beginning of the line

Frameworks/Cedar.xcframework/ios-arm64_x86_64-simulator/Cedar.framework/Headers/CDRSpec.h
```diff
@@ -16,24 +16,30 @@ extern const __nullable CDRSpecBlock PENDING;
 #ifdef __cplusplus
 extern "C" {
 #endif
-void beforeEach(CDRSpecBlock);
+
+    void beforeEach(CDRSpecBlock);
```
tatiana-yan: and here too

tatiana-yan (approved):

Frameworks/Cedar.xcframework/ios-arm64_x86_64-simulator/Cedar.framework/Headers/CDRSpec.h
aleksproger: done

Frameworks/Cedar.xcframework/ios-arm64/Cedar.framework/Headers/CDRSpec.h
aleksproger: done

